PR #48, opened 2026-04-06
[#43] Resizable drag handle between terminal and preview panels

### LINKED ISSUE #43: [Feature] Resizable drag handle between terminal and preview panels
## Problem
The terminal (Claude CLI) and content preview sections have fixed widths. Users cannot adjust the split to give more room to either panel.

## Solution
Add a draggable divider/handle between the terminal panel and the preview panel:

- Vertical drag handle (thin bar, ~4px, with a visible grip indicator)
- Mouse drag to resize — update panel widths in real-time
- Minimum width constraint on both panels (e.g., 300px each)
- Persist the user's preferred split ratio in localStorage
- On window resize, maintain the ratio
- Cursor changes to `col-resize` on hover

### Design
- Handle color: `var(--border)` with subtle hover highlight
- Grip dots or line in the center of the handle
- Consistent with Moleskine design

## Acceptance Criteria
- [ ] Drag handle visible between terminal and preview panels
- [ ] Dragging resizes both panels in real-time
- [ ] Minimum width enforced on both sides
- [ ] Split ratio persisted in localStorage
- [ ] Cursor shows `col-resize` on hover
- [ ] Works on desktop (mobile is single-column, not affected)

## Branch
`task/<issue>-resizable-panels`

realproject7:
## Summary
- Adds a draggable divider (6px handle with grip dots) between the terminal and preview panels
- Replaces fixed `flex-1`/`w-96` layout with ratio-based flex sizing
- Drag resizes both panels in real-time with 300px minimum enforced on each side
- Split ratio persisted to `localStorage` and restored on reload
- `col-resize` cursor on hover, Moleskine-style handle color (`var(--border)`)

## Test plan
- [ ] Drag handle visible between terminal and preview panels
- [ ] Dragging resizes both panels in real-time
- [ ] Minimum 300px width enforced on both sides
- [ ] Refresh page → split ratio restored from localStorage
- [ ] Cursor shows `col-resize` on hover over handle
- [ ] Grip dots visible centered on handle
- [ ] Other panel functionality (preview tabs, publish, story browser) unaffected

Fixes #43

### DIFF
--- a/app/web/components/StoriesPage.tsx
+++ b/app/web/components/StoriesPage.tsx
@@ -11,6 +11,8 @@ interface StoriesPageProps {
 const STORAGE_KEY = "plotlink-panel-ratio";
 const DEFAULT_RATIO = 0.6; // terminal gets 60% of available space
 const MIN_PANEL_PX = 300;
+const SIDEBAR_PX = 224; // w-56
+const HANDLE_PX = 6;
 
 function loadRatio(): number {
   try {
@@ -21,6 +23,14 @@ function loadRatio(): number {
     }
   } catch { /* ignore */ }
   return DEFAULT_RATIO;
+}
+
+function clampRatio(r: number, available: number): number {
+  if (available <= 0) return r;
+  const minR = MIN_PANEL_PX / available;
+  const maxR = 1 - MIN_PANEL_PX / available;
+  if (minR >= maxR) return 0.5; // panels can't both fit, split evenly
+  return Math.min(maxR, Math.max(minR, r));
 }
 
 export function StoriesPage({ token, authFetch }: StoriesPageProps) {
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The drag handle itself is straightforward, but the split-ratio math is currently inconsistent between drag-time calculations and the rendered layout. That causes the restored/default split to be visibly wrong and can violate the 300px minimum without the user dragging.

### Findings
- [high] The resize math stores `ratio` as a fraction of the width *after* subtracting the 224px sidebar, but the render applies that same ratio as `flex-basis` against the full container width. For example, at a 1200px container and the default `0.6` ratio, drag math intends a ~586px terminal, but the rendered terminal becomes 720px and the preview drops to about 250px. That means the default/restored layout can start below the required 300px minimum before the user drags at all.
  - File: `app/web/components/StoriesPage.tsx:57`
  - Suggestion: Apply the ratio to the resizable area only, not the full container, or restructure the layout so the resizable terminal/handle/preview region is its own flex container separate from the fixed sidebar.

- [medium] Minimum-width enforcement only happens inside `onMouseMove`. A ratio restored from `localStorage` is accepted as-is, and there is no reclamp on initial render or window resize. If a user saves a valid wide-screen split and later opens the page in a narrower desktop window, one panel can render below 300px until they drag again.
  - File: `app/web/components/StoriesPage.tsx:15`
  - Suggestion: Clamp the loaded/stored ratio against the current container width on mount and on resize, not only during active dragging.

### Decision
Requesting changes because the current implementation does not reliably preserve the intended split ratio and can violate the issue's minimum-width requirement in normal desktop use.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The rereview addresses both requested issues. The resize ratio is now applied consistently to the resizable region, and stored ratios are clamped on mount and window resize so the 300px minimum is enforced outside active dragging.

### Findings
- None.

### Decision
Approving because the updated `StoriesPage` now matches issue #43's intended resize behavior and I did not find a new regression in the latest patch.

PR #79, opened 2026-04-06
[#66] Retry publish: new on-chain tx with fresh IPFS upload

### LINKED ISSUE #66: [Feature] Retry publish: new on-chain tx with fresh IPFS upload
## Problem
When indexing fails after publish, the user needs a way to retry. Since the indexer has a 5-minute window, re-sending the index call won't work for old transactions.

## Solution
"Retry Publish" creates a brand new on-chain transaction:

1. Re-read the current file content from disk
2. Upload to IPFS (new CID)
3. Compute fresh content hash
4. Call `chainPlot` again with the same storylineId — contract allows multiple plots
5. Wait for confirmation
6. Call indexing API with `{ txHash, content }` (the fixed call from #62)
7. On success: update `.publish-status.json` with new txHash, CID, set `status: "published"`
8. On indexing failure again: set `status: "published-not-indexed"` (user can retry again)

### Cost
~$0.0005 gas per retry. Acceptable tradeoff vs building re-index infrastructure.

### Edge case
The contract will have two plots with the same content for this storyline. This is harmless — plotlink.xyz shows the latest indexed one. If this becomes a concern, a future cleanup could be added.

## Depends on
- plotlink-ows#64 (indexing failure status)
- plotlink-ows#65 (retry UI)

## Files
- `app/lib/publish.ts` — reuse existing `publishPlot()` flow
- `app/routes/publish.ts` — new endpoint or reuse existing publish route
- `app/web/components/PreviewPanel.tsx` — wire retry button

## Branch
`task/<issue>-retry-publish`

realproject7:
## Summary
- Add "Retry Publish" button in published-not-indexed action bar
- Triggers full re-publish flow: IPFS upload → new on-chain tx → indexing
- Uses existing `onPublish` callback (StoriesPage.handlePublish) — no new backend needed
- Two-tier retry UX: "Retry Index" (lightweight) and "Retry Publish" (full re-publish)
- Guidance text explains when to use which option

## Test plan
- [ ] published-not-indexed shows both "Retry Index" and "Retry Publish" buttons
- [ ] "Retry Publish" triggers full SSE publish flow with progress
- [ ] Successful re-publish → status updates to "published" (or "published-not-indexed" if index fails again)
- [ ] "Retry Index" still works independently
- [ ] Guidance text shown below buttons

Fixes #66

### DIFF
--- a/app/web/components/PreviewPanel.tsx
+++ b/app/web/components/PreviewPanel.tsx
@@ -274,13 +274,15 @@ export function PreviewPanel({ storyName, fileName, authFetch, onPublish, publis
               >
                 {retrying ? "Retrying..." : "Retry Index"}
               </button>
-              <button
-                onClick={() => storyName && fileName && onPublish?.(storyName, fileName)}
-                disabled={!!publishingFile}
-                className="px-3 py-1 border border-border text-xs rounded hover:bg-surface disabled:opacity-50"
-              >
-                {publishingFile === fileName ? "Publishing..." : "Retry Publish"}
-              </button>
+              {isPlot && (
+                <button
+                  onClick={() => storyName && fileName && onPublish?.(storyName, fileName)}
+                  disabled={!!publishingFile}
+                  className="px-3 py-1 border border-border text-xs rounded hover:bg-surface disabled:opacity-50"
+                >
+                  {publishingFile === fileName ? "Publishing..." : "Retry Publish"}
+                </button>
+              )}
               {fileData.txHash && (
                 <a
                   href={`https://basescan.org/tx/${fileData.txHash}`}


project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up closes the remaining scope mismatch in the retry-publish UI. `Retry Publish` is now limited to plot files, which matches issue #66's intended `chainPlot`-based retry behavior and avoids the duplicate-storyline problem on genesis files.

### Findings
- None.

### Decision
Approving because the current head now satisfies issue #66's retry-publish requirements without exposing the feature in cases where the underlying flow would do the wrong on-chain operation.

Issue #103, opened 2026-04-07
[Research] Root cause analysis: plot indexing failures + retry-induced index inflation


realproject7:
## Background
Plot publishing from plotlink-ows has had repeated indexing failures, requiring multiple "Retry Publish" attempts. Each retry creates a new on-chain `chainPlot` tx, which causes the on-chain `plotIndex` to inflate (e.g., plot-01.md ended up at index 5 instead of 1 after 4 failed retries).

While the stored `plotIndex` from the event is correct and stable, this causes:
1. **Cost waste**: Each retry costs gas (~$0.0005 per attempt)
2. **URL inconsistency**: `plot-01.md` has URL `/story/48/5` instead of `/story/48/1`
3. **Confusion**: Filename ↔ URL mismatch makes manual debugging harder
4. **Wasted on-chain state**: Duplicate plot entries in the contract

### Phase 1: Root cause analysis
Investigate WHY indexing fails so often. Possible causes to verify:

1. **Timing race**: plotlink-ows calls `/api/index/plot` immediately after tx confirmation, but plotlink.xyz's indexer may not see the tx yet (RPC propagation lag, block finality)
2. **IPFS propagation**: Filebase/IPFS may not have propagated the content CID yet when plotlink fetches it (10s timeout in indexer)
3. **Content hash format mismatch**: Even with the #62 fallback fix, hashes might still mismatch due to encoding (UTF-8 BOM, line endings, trailing whitespace)
4. **5-minute window**: Tx confirmation + IPFS + indexer call must happen within 5 min — if any step is slow, indexing window expires
5. **Retry-on-failure blind spots**: When retry-index fails (within 5 min), is the user notified? Or does it silently transition to "retry publish"?

### Phase 2: Alternative solutions
Propose safe, easy-to-implement alternatives that reduce or eliminate indexing failures. Examples:

1. **Wait + verify before declaring "not indexed"**:
   - After publish, poll plotlink.xyz's `/story/{id}` endpoint to verify the plot appears
   - Only show "not indexed" status after N seconds of polling
   - Avoids false negatives from race conditions

2. **Defer indexing call**:
   - Wait 5-10 seconds after tx confirmation before calling indexer
   - Gives RPC + IPFS time to propagate
   - Tradeoff: slower UX, but higher success rate

3. **Pre-warm IPFS content**:
   - Upload to IPFS, then immediately fetch the CID via the gateway to confirm propagation
[...]

PR #106, opened 2026-04-07
[#105] Add delay + retry for plotlink indexing

### LINKED ISSUE #105: [Feature] Plot indexing: delay + auto-retry loop (A+B from #103 research)
## Background
Per research findings in #103, the dominant cause of plot indexing failures is an RPC propagation race: plotlink-ows calls `/api/index/plot` immediately after `waitForTransactionReceipt` returns, but plotlink.xyz's indexer (on a different RPC provider) hasn't seen the tx yet. Users then click "Retry Publish" after the 5-min window closes on "Retry Index", which creates duplicate `chainPlot` txs and inflates the `plotIndex`.

## Solution: Combo A+B
Implement the recommended combo from #103 in `app/lib/publish.ts` — one function, no new endpoints, no plotlink.xyz changes.

### A. Defer indexing call by 8 seconds
Insert a fixed 8s delay after `waitForPlotConfirmation` and before the `/api/index/plot` POST. Eliminates RPC propagation race in the common case.

### B. Auto-retry loop
Wrap the indexer POST in a retry loop:
- Up to 10 attempts total
- 30s interval between attempts (after the initial 8s delay)
- Total window: ~4.5 minutes (within plotlink's 5-min recency check)
- Stream progress via existing SSE: `"Indexing… (attempt N/10)"` — keeps user informed so they don't escalate to Retry Publish
- On success: exit loop, mark as published
- On all 10 attempts failing: store `indexError`, show the existing Retry Index button (per #64/#65)

### Apply to both flows
- `publishStoryline` (lines 223-279) — storyline indexing
- `publishPlot` (lines 297-367) — plot indexing

Both have the same race condition; both benefit from the fix.

## Why this combo
- Fixes the dominant cause (RPC propagation) with a single `setTimeout`
- Closes the cost-waste loop: users can't trigger duplicate `chainPlot` txs until the *entire* 5-min window is exhausted
- Backward-compatible with existing stored-failure UX (#64, #65, #71)
- ~30 LOC change, concentrated in one function per publish flow
- No new endpoints, no auth changes, no plotlink.xyz coordination

## Acceptance Criteria
- [ ] 8s delay before initial indexer POST
- [ ] Auto-retry up to 10× at 30s intervals on failure
- [ ] Progress stream shows "Indexing… (attempt N/10)"
- [ ] Applied to both `publishStoryline` and `publishPlot`
- [ ] On success, normal published status
- [ ] On all retries failing, existing "Retry Index" UI kicks in
- [ ] Test with a real publish: verify the indexing lands on first or second attempt

## Files
- `app/lib/publish.ts` — add `indexWithRetry()` helper, use in both publish flows
- `app/routes/publish.ts` — pass retry progress through SSE stream

## Branch
`task/<issue>-indexing-delay-retry`

realproject7:
Fixes #105 (implements #103 research recommendation)

## Summary
- Add `indexWithDelayAndRetry` helper in `app/lib/publish.ts`
- 8s initial delay before the first `/api/index/{plot,storyline}` POST (combo A — RCA #1: RPC propagation race)
- 10× 30s retry loop, total ~4.5 min (combo B — RCA #5: prevents Retry Publish escalation that inflates on-chain plotIndex)
- New `indexing` step in `PublishProgress` union, streams `Indexing… (attempt N/10)` per attempt
- Applied to both `publishStoryline` and `publishPlot`

## Why
Per the #103 research comment, the dominant indexing-failure mode is the race between viem's 1-confirmation `waitForTransactionReceipt` returning and plotlink.xyz's indexer actually seeing the tx on its RPC. The 8s delay closes that race in the common case, and the 10×30s retry loop covers the long-tail (IPFS propagation, transient indexer errors) without burning a second `chainPlot` tx.

## Test plan
- [ ] Type check passes (`npx tsc --noEmit` ✅ verified locally)
- [ ] Publish a fresh storyline and verify the SSE progress stream shows `indexing` events with `Indexing… (attempt N/10)` messages
- [ ] Publish a plot to an existing storyline and verify same progress stream
- [ ] Confirm `indexError` is undefined on success and contains the final error after all 10 attempts on failure
- [ ] (Manual) Force a slow indexer (block plotlink.xyz on the network for 60s, then unblock) and verify the retry loop recovers without the user clicking Retry Publish

### DIFF
--- a/app/lib/publish.ts
+++ b/app/lib/publish.ts
@@ -61,12 +61,81 @@ export interface PublishResult {
 }
 
 export interface PublishProgress {
-  step: "uploading" | "estimating" | "signing" | "broadcasting" | "confirming" | "done" | "error";
+  step: "uploading" | "estimating" | "signing" | "broadcasting" | "confirming" | "indexing" | "done" | "error";
   message: string;
   txHash?: string;
   contentCid?: string;
   storylineId?: number;
   error?: string;
+}
+
+// Indexing retry tuning (per #103 RCA — combo A+B):
+// - 8s initial delay lets the on-chain tx propagate to plotlink.xyz's RPC
+//   and gives Filebase → public IPFS gateway propagation a head start.
+// - 10 attempts × 30s interval ≈ 4.5 min total, fitting inside plotlink.xyz's
+//   5-min indexable window so users don't escalate to a full Retry Publish
+//   (which would mint another chainPlot tx and inflate the on-chain index).
+const INDEX_INITIAL_DELAY_MS = 8_000;
+const INDEX_RETRY_ATTEMPTS = 10;
+const INDEX_RETRY_INTERVAL_MS = 30_000;
+
+/**
+ * POST to plotlink.xyz's indexer with an initial delay and a retry loop.
+ * Streams "Indexing… (attempt N/M)" progress so the publish flow surfaces
+ * the in-progress state instead of an immediate failure.
+ *
+ * Returns undefined on success, or the final error message after all
+ * attempts have failed.
+ */
+async function indexWithDelayAndRetry(
+  endpoint: "plot" | "storyline",
+  body: Record<string, unknown>,
+  onProgress: (progress: PublishProgress) => void,
+  txHash: string,
+  contentCid: string,
+): Promise<string | undefined> {
+  const PLOTLINK_URL = process.env.NEXT_PUBLIC_APP_URL || "https://plotlink.xyz";
+  const url = `${PLOTLINK_URL}/api/index/${endpoint}`;
+
+  onProgress({
+    step: "indexing",
+    message: `Indexing… waiting ${INDEX_INITIAL_DELAY_MS / 1000}s for on-chain propagation`,
+    txHash,
+    contentCid,
+  });
+  await new Promise((r) => setTimeout(r, INDEX_INITIAL_DELAY_MS));
+
+  let lastError: string | undefined;
+  for (let attempt = 1; attempt <= INDEX_RETRY_ATTEMPTS; attempt++) {
+    onProgress({
+      step: "indexing",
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The retry helper in `app/lib/publish.ts` looks directionally right, but the PR does not yet satisfy the user-facing progress requirement from issue #105. The SSE consumer still renders only `data.step`, so users will see a generic `indexing` state instead of the required `Indexing… (attempt N/10)` progress.

### Findings
- [medium] The publish UI still discards the streamed progress message, so the retry loop does not surface the required attempt counter to users.
  File: `app/web/components/StoriesPage.tsx:191`
  Suggestion: when consuming the SSE events, prefer `data.message` over `data.step` (for example `setPublishProgress(data.message ?? data.step)`). Right now this PR streams `message: "Indexing… (attempt N/10)"`, but the UI only stores `data.step`, so the status bar will show just `indexing` and users still won't see attempt progress during the 4.5-minute retry window.

### Decision
Request changes. Please wire the existing SSE consumer to display the streamed `message` text so the new delay/retry flow actually meets #105's progress-stream acceptance criteria, then re-request review.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit resolves the only blocking issue from my prior review by wiring the SSE consumer to display the streamed progress message text. The retry/delay implementation remains scoped to `app/lib/publish.ts`, applies to both publish flows, and now actually surfaces the required `Indexing… (attempt N/10)` progress to users.

### Findings
- None.

### Decision
Approve. The PR now satisfies issue #105's repo-scoped acceptance criteria, with the remaining risk limited to runtime behavior validation in the pending checks/manual publish flow.

PR #111, opened 2026-04-12
[#108] Preserve owner identity when linking an OWS writer

### LINKED ISSUE #108: [Bug] Preserve owner identity when linking an OWS writer
## Scope
Audit the shared OWS user/agent profile flow so linking an OWS writer to a human owner does not collapse the owner's profile identity into the AI Writer identity.

This should preserve the distinction between:
- the human owner account
- the linked OWS writer wallet / agent identity

## Acceptance Criteria
- OWS onboarding / agent caching does not overwrite or conflate the human owner's identity with the linked writer identity
- Shared profile data keeps `agent_owner`, `agent_wallet`, and the human user's primary identity semantically distinct
- Any OWS profile surfaces that render agent identity use owner-aware labeling or linking rather than presenting the owner as the AI Writer itself
- The behavior stays aligned with PlotLink so both apps represent linked AI writers consistently

realproject7:
Fixes #108

## Summary
- **Onboarding guard**: `onboard/route.ts` no longer stamps agent columns (agent_id, agent_name, etc.) onto the human owner's user row when the agent has a separate bound wallet. Previously, an owner visiting their profile would trigger ERC-8004 detection that overwrote their identity with agent data.
- **fetchAgentMetadata filter**: Returns `null` when the lookup address is only the `agent_owner` (not the `agent_wallet`), preventing the profile page from displaying "AI Agent" badge and agent name/bio for human owners.
- **Linked AI Writer card**: New `getLinkedAgent()` server action returns minimal agent info (id, name, wallet) for owner addresses. Profile header renders a "Linked AI Writer" card with a link to the agent's profile.
- **detectWriterType tightened**: DB check now matches `agent_wallet` only (not `primary_address`), preventing false agent classification when an owner's row was previously corrupted with agent_id.
- Patch version bump: 0.1.20 → 0.1.21

## Test plan
- [ ] Visit `/profile/[owner_address]` where owner has a separate agent wallet — should show Human badge, owner's Farcaster identity, and "Linked AI Writer" card
- [ ] Visit `/profile/[agent_wallet]` — should show AI Agent badge, agent name/bio, and "Owner:" link as before
- [ ] Visit `/profile/[owner_address]` where owner IS the agent wallet (no separate binding) — should show AI Agent badge as before
- [ ] Trigger profile refresh (onboard) for an owner with separate agent wallet — agent columns should NOT appear on owner's row
- [ ] Create a storyline with agent wallet — `writer_type` should be 1 (agent)
- [ ] Create a storyline with owner wallet — `writer_type` should be 0 (human)

### DIFF
--- a/lib/actions.ts
+++ b/lib/actions.ts
@@ -48,6 +48,17 @@ export async function fetchAgentMetadata(
     ? preloadedUser
     : await getAgentUserFromDB(address);
   if (dbUser?.agent_id != null) {
+    const normalized = address.toLowerCase();
+    const agentWallet = dbUser.agent_wallet?.toLowerCase() ?? null;
+    const agentOwner = dbUser.agent_owner?.toLowerCase() ?? null;
+
+    // Don't return agent metadata when the address is only the owner and a
+    // separate agent_wallet exists. The owner is a human — their profile
+    // should not be presented as the AI Writer.
+    if (agentWallet && agentWallet !== normalized && agentOwner === normalized) {
+      return null;
+    }
+
     return {
       agentId: String(dbUser.agent_id),
       owner: dbUser.agent_owner ?? undefined,
@@ -91,8 +102,47 @@ export async function fetchAgentMetadata(
 }
 
 /**
+ * Minimal info about a linked agent, returned for owner profiles.
+ */
+export type LinkedAgent = {
+  agentId: string;
+  name: string;
+  agentWallet: string;
+};
+
+/**
+ * If `address` is an agent owner with a separate bound wallet, return
+ * minimal info so the profile page can link to the agent's profile.
+ */
+export async function getLinkedAgent(
+  address: string,
+): Promise<LinkedAgent | null> {
+  const supabase = createServiceRoleClient();
+  if (!supabase) return null;
+
+  const normalized = address.toLowerCase();
+  const { data } = await supabase
+    .from("users")
+    .select("agent_id, agent_name, agent_wallet")
+    .eq("agent_owner", normalized)
+    .not("agent_id", "is", null)
+    .not("agent_wallet", "is", null)
+    .single();
+
+  if (!data?.agent_wallet || data.agent_wallet.toLowerCase() === normalized) {
+    return null;
+  }
+
+  return {
+    agentId: String(data.agent_id),
+    name: data.agent_name ?? "Unknown Agent",
+    agentWallet: data.agent_wallet,
+  };
+}
+
+/**
  * Fetch the full user profile in a single DB lookup.
- * Returns dbUser, fcProfile, and agentMeta derived from one shared row.
+ * Returns dbUser, fcProfile, agentMeta, and linkedAgent derived from one shared row.
  * External API fallbacks still fire when DB data is missing.
  */
 export async function getFullUserProfile(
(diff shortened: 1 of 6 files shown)


realproject7 (commented):
**re2: APPROVE** ✓

Reviewed all 5 files. The identity separation logic is correct:

1. **`fetchAgentMetadata` guard** (lib/actions.ts) — Correctly returns `null` when the queried address matches `agent_owner` but not `agent_wallet`. This prevents the owner profile from rendering as AI Agent.

2. **`getLinkedAgent`** (lib/actions.ts) — Queries the `users` table by `agent_owner` to find the agent row (keyed by `agent_wallet`). This works because the agent wallet's row still gets `agent_owner` stamped during its own onboarding (since `hasSeparateWallet` is false when the address IS the agent wallet). Note: `.single()` will return null if an owner has multiple agents — acceptable for now.

3. **`detectWriterType`** (lib/contracts/erc8004.ts) — Narrowed from `.or(agent_wallet, primary_address)` to `.eq("agent_wallet")` only. Correct fix — prevents false agent classification from corrupted owner rows.

4. **Onboarding guard** (onboard/route.ts) — The `hasSeparateWallet` check correctly distinguishes: agent-wallet-onboard (stamps agent data) vs owner-wallet-onboard (skips). The logic: `boundWallet !== normalizedAddress` is the key discriminator.

5. **Profile UI** (page.tsx) — Linked AI Writer card rendered only when `!isAgent && linkedAgent`. Clean integration with existing ProfileHeader props.

Version bump 0.1.20 → 0.1.21 is correct per patch rules. No runtime regressions expected.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The profile rendering changes move in the right direction, but the onboarding persistence path still conflates the human owner and the linked agent into the same `users` row when the owner does not already have a separate row. That violates the issue's acceptance criteria about keeping owner identity and linked writer identity semantically distinct.

### Findings
- [high] Owner refresh can still overwrite the linked agent row instead of creating/updating a distinct human row.
  - File: `src/app/api/user/onboard/route.ts:41`
  - Suggestion: Do not treat a row found only via `agent_owner` as the owner's canonical user row for onboarding/upsert. Restrict the existing-user lookup for owner refreshes to human identity columns (`verified_addresses` / `primary_address`), or otherwise split the owner and agent persistence paths so refreshing the owner cannot update the agent row by id.
  - File: `lib/user-upsert.ts:45`
  - Suggestion: `findUserByWallet()` currently returns a row matched by `agent_owner`, and `upsertUser()` then updates that row in place. For an owner with a separate bound wallet and no standalone human row yet, this preserves the same shared row the bug is supposed to eliminate.

### Decision
Requesting changes because the PR still allows onboarding/agent caching to conflate owner and writer identity at the storage layer, even though the profile page now hides some of that conflation at render time.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up change fixes the storage-layer identity conflation I flagged by preventing owner onboarding from resolving an existing row through `agent_owner`. With that in place, the PR now preserves the distinction between the human owner profile and the linked AI writer identity across persistence, detection, and profile rendering.

### Findings
- None.

### Decision
Approved on code review. The `skipAgentOwner` path in onboarding closes the remaining correctness gap from my prior review. Note: `gh pr checks` currently reports no checks on this branch, so merge readiness should be confirmed separately if CI is expected.

realproject7 (commented):
**re2: APPROVE (re-review)** ✓

RE1's finding was valid — I missed the `findUserByWallet()` → `agent_owner` lookup path that could return the agent's row when an owner onboards.

**Fix verified:** `skipAgentOwner` option in `findUserByWallet()` (lib/user-upsert.ts:46) short-circuits the lookup chain before the `agent_owner` query. The onboard route now passes `{ skipAgentOwner: true }` (route.ts:40), so a human owner's refresh never returns or updates the linked agent's row.

The rest of the PR (fetchAgentMetadata guard, getLinkedAgent, detectWriterType, profile UI) remains correct from my initial review. Storage-layer gap is now closed.

PR #135, opened 2026-04-23
[#134] Move user data to ~/.plotlink-ows/ for npx persistence

realproject7:
## Summary
- **Fixes #134** — stories and database stored in package dir were lost on `npx plotlink-ows@latest` upgrades
- All user data (`stories/`, `data/local.db`, terminal sessions) now lives in `~/.plotlink-ows/` which persists across npx reinstalls
- Auto-migrates existing data from old package-relative paths on first startup

## Changes
- `app/lib/paths.ts` — Added `STORIES_DIR`, `DATA_DIR`, `DB_PATH`, `DATABASE_URL` constants pointing to `~/.plotlink-ows/`
- `app/server.ts` — Uses persistent paths; sets `DATABASE_URL` env before Prisma; adds `migrateOldData()` on startup
- `app/routes/stories.ts` — Imports `STORIES_DIR` from `paths.ts` instead of hardcoding `__dirname` relative path
- `app/routes/terminal.ts` — Imports `STORIES_DIR` and `DATA_DIR` from `paths.ts`
- `app/prisma/schema.prisma` — Uses `env("DATABASE_URL")` instead of hardcoded relative path
- `scripts/fix-index-status.ts` — Imports `STORIES_DIR` from `paths.ts`

## Test plan
- [ ] `npm run lint` — passes (no new errors)
- [ ] `npm run typecheck` — passes
- [ ] `npm run build` — passes
- [ ] `npm run app:build` — passes
- [ ] Verify `~/.plotlink-ows/stories/` and `~/.plotlink-ows/data/` are created on startup
- [ ] Verify stories survive simulated npx reinstall (different `__dirname`)
- [ ] Verify auto-migration moves old data from package-relative paths

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #139, opened 2026-04-23
[#136] Scan npx cache dirs for story auto-migration

### LINKED ISSUE #136: [Bug] Auto-migration fails to find stories from previous npx install
## Problem

PR #135 added auto-migration from package-relative paths to `~/.plotlink-ows/`. But the migration only checks `path.join(__dirname, "..", "stories")` — which points to the **current** npx cache directory, not the previous one.

When npx installs a new version, it creates a fresh cache directory. The old stories are in `~/.npm/_npx/<old-hash>/...` but the migration looks in `~/.npm/_npx/<new-hash>/...` which is empty.

**Result:** Stories are still lost on upgrade despite the migration code.

## Root Cause

`__dirname` always resolves to the current install path. There's no way to know the previous npx cache hash.

## Fix

On startup, scan `~/.npm/_npx/*/node_modules/plotlink-ows/stories/` for any existing story directories and migrate them to `~/.plotlink-ows/stories/`. This catches stories from ANY previous npx install.

```typescript
import { globSync } from "fs";

function migrateFromNpxCaches() {
  const npxBase = path.join(os.homedir(), ".npm", "_npx");
  if (!fs.existsSync(npxBase)) return;
  
  for (const hash of fs.readdirSync(npxBase)) {
    const oldStories = path.join(npxBase, hash, "node_modules", "plotlink-ows", "stories");
    if (!fs.existsSync(oldStories)) continue;
    
    const entries = fs.readdirSync(oldStories, { withFileTypes: true })
      .filter(d => d.isDirectory() && !d.name.startsWith(".") && d.name !== "_example");
    
    for (const entry of entries) {
      const dest = path.join(STORIES_DIR, entry.name);
      if (!fs.existsSync(dest)) {
        fs.cpSync(path.join(oldStories, entry.name), dest, { recursive: true });
        console.log(`  Migrated story "${entry.name}" from npx cache`);
      }
    }
  }
}
```

Also migrate from dev directory path (`Projects/plotlink-ows/stories/`) as a secondary check.

## Acceptance Criteria

- [ ] Stories from any previous npx cache are auto-migrated on startup
- [ ] No duplicate migration (skip if already exists in destination)
- [ ] Works for both npx upgrades and dev→npx transitions

realproject7:
## Summary
- **Fixes #136** — auto-migration from #135 only checked current `__dirname`, missing stories from previous npx installs
- Now scans `~/.npm/_npx/*/node_modules/plotlink-ows/stories/` to find stories from ANY prior npx cache
- Also migrates database and terminal sessions from old npx caches
- Keeps dev→npx migration path as secondary fallback
- Refactored migration into `migrateStoriesFrom()` and `migrateFileFrom()` helpers to avoid duplication

## Changes
- `app/server.ts` — Rewrote `migrateOldData()` to scan all npx cache directories, extracted reusable helpers
- `package.json` — Version bump 1.0.5 → 1.0.6

## Test plan
- [ ] `npm run typecheck` — passes
- [ ] `npm run lint` — no new errors
- [ ] Verify stories from old npx caches are found and migrated on startup
- [ ] Verify duplicate stories are skipped (no overwrite)
- [ ] Verify dev→npx transition still works

### DIFF
--- a/app/server.ts
+++ b/app/server.ts
@@ -1,4 +1,5 @@
 import dotenv from "dotenv";
+import os from "os";
 import path from "path";
 import { fileURLToPath } from "url";
 import { ENV_FILE, DATA_DIR, STORIES_DIR, DATABASE_URL } from "./lib/paths";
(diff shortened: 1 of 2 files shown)


realproject7:
## Review — @re2

### ✅ APPROVED

Clean implementation:
- **npx cache scan** — iterates `~/.npm/_npx/*/node_modules/plotlink-ows/` to find stories from any prior npx install. Correctly best-effort with try/catch around the scan.
- **Refactored helpers** — `migrateStoriesFrom` and `migrateFileFrom` reduce duplication and keep skip-if-exists + EXDEV fallback logic from #135.
- **Dual migration path** — npx cache scan first, then package-relative as secondary (dev→npx transition). Order is correct since `migrateFileFrom` skips if dest already exists.
- **Version** — 1.0.6, correct patch bump.

CI pending at time of review — no code concerns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The story-directory scan matches issue #136, but the same loop now also migrates `data/local.db` and `terminal-sessions.json` from an arbitrary npx cache entry. With multiple prior npx installs, that can restore stale state from an older cache instead of the most recent one.

### Findings
- [high] Database/session migration is nondeterministic across multiple npx caches.
  - File: `app/server.ts:95`
  - Suggestion: limit the new cache scan to story directories only, or deterministically choose the newest valid cache before copying singleton files like `local.db` and `terminal-sessions.json`. The current `fs.readdirSync(npxBase)` order is not a safe source-of-truth for singleton state.

### Decision
Request changes. The PR fixes story discovery, but the added database/session migration path can resurrect stale user state and needs to be made deterministic or removed from this scan.

realproject7:
## Re-review — @re2

### ✅ APPROVED

Fix correctly addresses @re1's concern: npx cache scan is now stories-only (additive, skip-duplicates is safe for multi-cache). DB and terminal sessions only migrate from the deterministic `__dirname`-relative path. No new issues.

project7-interns (commented):
## Verdict: COMMENT

### Summary
The singleton-state regression is fixed: the npx cache scan now migrates only story directories, and database/session migration is limited to the deterministic package-relative fallback path.

### Findings
- No remaining code blockers in the updated diff.
- `lint-and-typecheck` is still pending on the current head commit, so build status is not yet confirmed.

### Decision
No code blockers from re1 on the updated patch. I am holding final approval until the current PR checks finish successfully.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete on the latest head commit. The earlier migration issue is fixed, and the current `lint-and-typecheck` run is now passing.

### Findings
- No blocking findings.

### Decision
Approve. PR #139 now has the required re1 approval on the current passing revision.

PR #164, opened 2026-04-24
Fix: Reuse untitled session when new story is created

### LINKED ISSUE #162: [Bug] New story creates separate terminal session — should reuse the Untitled session
## Problem

When a user clicks "+ New Story" and starts chatting with Claude, the AI creates a new story folder (e.g., \`paper-chair\`). But instead of the Untitled session becoming the \`paper-chair\` session, a **new separate terminal session** is spawned for \`paper-chair\`. The original Untitled session is left orphaned with a separate Claude instance.

This means:
- Two Claude sessions running (wasted resources)
- The context from the Untitled session (where the user discussed the story idea) is lost in the new session
- The new \`paper-chair\` session starts fresh without the brainstorming context

## Expected Flow

1. **+ New Story** → launches Untitled terminal session (Claude CLI in \`~/.plotlink-ows/stories/\`)
2. User chats with Claude, brainstorms ideas
3. Claude creates the story folder (e.g., \`mkdir stories/paper-chair\` + creates \`structure.md\`)
4. **The Untitled session seamlessly becomes the \`paper-chair\` session** — same terminal, same Claude instance, same conversation context
5. Sidebar updates: "Untitled" → "Paper Chair" (from structure.md title)
6. Terminal tab updates: "Untitled" → "paper-chair"

## Current Behavior

- Step 3 happens: Claude creates the folder
- The polling detects the new folder → creates a NEW terminal session for \`paper-chair\`
- The Untitled session remains as a separate orphaned session
- User now has two tabs: "Untitled" (old) + "paper-chair" (new, no context)

## Fix

When the polling in \`StoriesPage.tsx\` detects a new story folder:
1. **Don't spawn a new terminal session** for it
2. Instead, **rename the oldest Untitled session** to the new story name
3. Update the PTY session key in the \`ptySessions\` Map (backend) from \`_new_XXX\` to \`paper-chair\`
4. Update the terminal tab display name
5. The Claude CLI process continues running — same PID, same conversation

### Backend change needed

Add an endpoint to rename a session key:
\`POST /api/terminal/rename\` — body: \`{ oldName: "_new_1234", newName: "paper-chair" }\`

This moves the PTY entry in the sessions Map without killing/restarting the process.

### Frontend change needed

In the polling effect (\`StoriesPage.tsx\`), when a new story is detected:
- Call the rename endpoint instead of letting a new session auto-create
- Update the session key in the local Map
- Remove the Untitled entry from \`untitledSessions\` state

## Files

- \`app/web/components/StoriesPage.tsx\` — polling logic for new story detection
- \`app/web/components/TerminalPanel.tsx\` — session key management
- \`app/routes/terminal.ts\` — add rename endpoint, update session Map

## Acceptance Criteria

- [ ] New story creation reuses the Untitled terminal session
- [ ] Claude conversation context preserved (no restart)
- [ ] Sidebar updates from "Untitled" to story title
- [ ] Terminal tab updates from "Untitled" to story name
- [ ] No orphaned Untitled sessions after story creation
- [ ] Only one Claude process running per story

realproject7:
## Summary
- Adds `POST /api/terminal/rename` backend endpoint that moves a PTY session entry to a new key without killing the process
- Frontend polling now calls rename when a new story folder is detected, instead of letting a new session auto-create
- Preserves Claude conversation context — no orphaned untitled sessions

Fixes #162

## Test plan
- [ ] Click "+ New Story", brainstorm with Claude until it creates a story folder
- [ ] Verify the Untitled tab renames to the story name (no new tab spawned)
- [ ] Verify Claude conversation context is preserved (same session, same PID)
- [ ] Verify no orphaned Untitled sessions remain
- [ ] Verify closing and reopening the story tab still resumes correctly

### DIFF
--- a/app/routes/terminal.ts
+++ b/app/routes/terminal.ts
@@ -84,14 +84,19 @@ function spawnPty(storyName: string, opts?: { sessionId?: string; resume?: boole
   ptySessions.set(storyName, session);
 
   term.onExit(({ exitCode }) => {
-    const s = ptySessions.get(storyName);
-    if (s?.term !== term) return;
+    // Find this session by term reference — key may have changed via rename
+    let currentName: string | undefined;
+    let s: typeof session | undefined;
+    for (const [key, entry] of ptySessions) {
+      if (entry.term === term) { currentName = key; s = entry; break; }
+    }
+    if (!currentName || !s) return;
 
     // If a resumed session exits quickly (< 5s), signal client to auto-reconnect fresh
     const elapsed = Date.now() - spawnTime;
     if (isResume && elapsed < 5000 && exitCode !== 0) {
-      console.log(`Resume for "${storyName}" failed (exit ${exitCode} in ${elapsed}ms), signaling fresh fallback`);
-      ptySessions.delete(storyName);
+      console.log(`Resume for "${currentName}" failed (exit ${exitCode} in ${elapsed}ms), signaling fresh fallback`);
+      ptySessions.delete(currentName);
       if (s.ws && s.ws.readyState <= 1) {
         // Close code 4000 = resume-failed, client should auto-reconnect fresh
         s.ws.close(4000, "resume-failed");
--- a/app/routes/terminal.ts
+++ b/app/routes/terminal.ts
@@ -193,6 +193,30 @@ terminal.delete("/:storyName/discard", (c) => {
   return c.json({ ok: true });
 });
 
+/** POST /api/terminal/rename — rename a session key without killing the process */
+terminal.post("/rename", async (c) => {
+  const body = await c.req.json<{ oldName?: string; newName?: string }>().catch(() => ({}));
+  const oldName = body.oldName && safeName(body.oldName);
+  const newName = body.newName && safeName(body.newName);
+  if (!oldName || !newName) return c.json({ error: "Invalid names" }, 400);
+  if (oldName === newName) return c.json({ ok: true });
+
+  const session = ptySessions.get(oldName);
+  if (!session) return c.json({ error: "Session not found" }, 404);
+
+  // Move in-memory PTY entry
+  ptySessions.delete(oldName);
+  ptySessions.set(newName, session);
+
+  // Update persisted session map: remove old key, store under new key
+  const sessionMap = loadSessionMap();
+  delete sessionMap[oldName];
+  sessionMap[newName] = session.sessionId;
+  saveSessionMap(sessionMap);
+
+  return c.json({ ok: true, sessionId: session.sessionId });
+});
+
 /** POST /api/terminal/stop — kill PTY (legacy, kills default) */
 terminal.post("/stop", (c) => {
   const session = ptySessions.get("default");
--- a/app/routes/terminal.ts
+++ b/app/routes/terminal.ts
@@ -209,6 +209,8 @@ terminal.post("/rename", async (c) => {
   const session = ptySessions.get(oldName);
   if (!session) return c.json({ error: "Session not found" }, 404);
 
+  if (ptySessions.has(newName)) return c.json({ error: "Target session already exists" }, 409);
+
   // Move in-memory PTY entry
   ptySessions.delete(oldName);
   ptySessions.set(newName, session);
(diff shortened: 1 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR is close, but the rename flow still has two blocking session-lifecycle bugs. One leaves renamed PTY entries stuck in the active-session map after the process exits, and the other still allows a second session to spawn if the async rename does not complete before selection changes.

### Findings
- [high] Renaming the `ptySessions` map key does not update the exit handler's lookup key, so a renamed session is no longer cleaned up when the PTY exits.
  - File: `app/routes/terminal.ts:203`
  - Suggestion: store the current story/session key on the session object or otherwise make the `term.onExit` cleanup follow the renamed key before moving the map entry.
- [high] The new-story polling path fires `renameRef.current(oldName, name)` without awaiting success, then immediately removes the untitled entry and selects the new story.
  - File: `app/web/components/StoriesPage.tsx:90`
  - Suggestion: await the rename result and only switch selection / consume the untitled session after the rename succeeds; otherwise a slow or failed rename can still auto-spawn a second `paper-chair` session while the original untitled PTY keeps running.

### Decision
Requesting changes because the current implementation can still orphan or duplicate sessions, which violates the issue's acceptance criteria around reusing the original Untitled PTY.

realproject7:
## @re2 Review — REQUEST CHANGES

Overall the approach is clean — rename-in-place is the right call for preserving the PTY process and Claude conversation context. One issue needs fixing:

### 🔴 Missing server-side guard for existing `newName` (terminal.ts)

The frontend checks `sessions.has(newName)` before calling rename, but the backend endpoint does not check `ptySessions.has(newName)`. If two requests race or a client bypasses the check, the rename would silently overwrite an existing live session — orphaning its PTY process and WebSocket.

**Fix:** Add a guard after the `session` lookup:

```ts
if (ptySessions.has(newName)) return c.json({ error: "Target session already exists" }, 409);
```

Same applies to the session map — if `sessionMap[newName]` already exists, the old entry gets lost.

### ✅ Everything else looks good

- `safeName()` validation on both inputs — correct
- No-op on `oldName === newName` — good edge case handling
- Scrollback migration with cleanup of old key — solid
- `disconnected` Set update in React state — correct
- `useCallback` deps and ref cleanup in effect — clean
- Patch version bump — appropriate

One fix, then APPROVE.

project7-interns (commented):
## Verdict: REQUEST CHANGES

### Summary
The original rename-lifecycle bugs are fixed on the current head, but one backend blocker remains in the rename endpoint.

### Findings
- [high] `POST /api/terminal/rename` still does not reject `newName` when that target session key already exists in `ptySessions`, so a duplicate/racy rename can overwrite a live session entry and orphan its PTY.
  - File: `app/routes/terminal.ts:206`
  - Suggestion: add an explicit guard before moving the map entry, e.g. return `409` when `ptySessions.has(newName)`.

### Decision
Keeping this in request-changes state until the rename endpoint protects against collisions. The earlier cleanup/await fixes look correct.

project7-interns (commented):
## Verdict: COMMENT

### Summary
The current head addresses the remaining rename-collision blocker. I found no further code issues in the rename/session-reuse flow.

### Findings
- No remaining code blockers on the current diff.

### Decision
Holding final approval until the current `lint-and-typecheck` run is green.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The rename/session-reuse implementation now satisfies issue #162. The current head fixes the session cleanup, async rename sequencing, and rename-collision handling.

### Findings
- No remaining blocking issues.

### Decision
Approving because the implementation meets the acceptance criteria and `lint-and-typecheck` passed on the current head.

realproject7:
## @re2 Re-review — APPROVE ✅

All issues addressed across the three commits:

1. **409 collision guard** (my original request) — `ptySessions.has(newName)` check now in place, returns 409. Good.
2. **onExit handler fix** (@re1's catch) — now finds the session by `term` reference instead of the captured closure name. Correct — rename no longer breaks cleanup.
3. **Await rename + boolean return** — `StoriesPage` now awaits the rename and only consumes the untitled session on success. Failed renames leave state intact. Solid error handling.

CI (`lint-and-typecheck`) is passing. No remaining blockers from my side.

**APPROVE**

PR #238, opened 2026-05-28
[#237] Wire cartoon readiness checks into publish blocking

realproject7:
## Summary

- **Server defense**: `POST /api/publish/file` runs `checkMarkdownReadiness` for cartoon plot files (reads `plot-NN.cuts.json` + markdown). Returns 400 + issues when marker blocks missing, awaiting/pending placeholders present, or image refs aren't uploaded http(s)/IPFS URLs. Missing/invalid cuts.json also blocks. Direct API calls cannot bypass.
- **UI blocking**: PreviewPanel computes cartoon publish readiness (fetches cuts + content), disables the Publish button and shows actionable issues near it when not ready.
- **Policy**: every publishable cartoon cut must become a final uploaded image — removed the narration-only null-image-publish guidance from instructions.
- Fiction and genesis publish unchanged.

## Test plan

- [ ] `npm run typecheck` passes
- [ ] `npm run lint` passes (0 errors)
- [ ] `npm run test` passes (288 tests)
- [ ] `npm run app:build` passes
- [ ] Server: cartoon plot with placeholder/local-path/missing-marker/invalid-cuts → 400 with issues
- [ ] Server: fiction plot not subject to cartoon guard
- [ ] UI: cartoon plot with unready markdown disables Publish + shows issues
- [ ] UI: ready cartoon markdown shows no issues
- [ ] Valid uploaded cartoon markdown can publish
- [ ] Fiction plot publish unchanged

Closes #237

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #240, opened 2026-05-28
[#239] Verify cartoon markdown image URLs match uploaded cut URLs

realproject7:
## Summary

- Strengthen `checkMarkdownReadiness` to tie publish markdown to canonical uploaded cut state
- Per cut: require `cut.uploadedUrl` is set; extract the `ows:cartoon-cut` block; require exactly one image reference; require that image URL exactly equals `cut.uploadedUrl`
- Reject arbitrary `https://` URLs that aren't the recorded upload, blocks with no image, multi-image blocks
- Kept from #233/#237: local/relative ref blocking and pending-placeholder blocking
- Closes the gap where a hand-written arbitrary https image could pass readiness even though the cut was never uploaded through OWS
- Fiction publish behavior unchanged

## Test plan

- [ ] `npm run typecheck` passes
- [ ] `npm run lint` passes (0 errors)
- [ ] `npm run test` passes (298 tests)
- [ ] Cut with `uploadedUrl: null` fails readiness even with valid-looking URL
- [ ] Mismatched URL fails (`example.com/fake` vs recorded `ipfs/QmA`)
- [ ] Exact recorded URL passes
- [ ] No-image and multi-image blocks fail
- [ ] Server `/api/publish/file` returns 400 for mismatch/null, allows exact match
- [ ] Generated markdown still passes (matches uploadedUrl by construction)

Closes #239

### DIFF
--- a/app/lib/cartoon-readiness.test.ts
+++ b/app/lib/cartoon-readiness.test.ts
@@ -134,6 +134,63 @@ describe("checkMarkdownReadiness", () => {
     const { ready } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: "https://ipfs.example.com/QmAbc" })]);
     expect(ready).toBe(true);
   });
+
+  it("fails when cut.uploadedUrl is null even with a valid-looking URL in markdown", () => {
+    const md = [
+      "<!-- ows:cartoon-cut cut-001 start -->",
+      "![Cut 1](https://ipfs.filebase.io/ipfs/QmReal)",
+      "<!-- ows:cartoon-cut cut-001 end -->",
+    ].join("\n");
+    const { ready, issues } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: null })]);
+    expect(ready).toBe(false);
+    expect(issues.some((i) => i.includes("not uploaded"))).toBe(true);
+  });
+
+  it("fails when block URL does not match recorded uploadedUrl", () => {
+    const md = [
+      "<!-- ows:cartoon-cut cut-001 start -->",
+      "![Cut 1](https://example.com/fake.webp)",
+      "<!-- ows:cartoon-cut cut-001 end -->",
+    ].join("\n");
+    const { ready, issues } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: "https://ipfs.filebase.io/ipfs/QmA" })]);
+    expect(ready).toBe(false);
+    expect(issues.some((i) => i.includes("does not match the recorded uploaded URL"))).toBe(true);
+  });
+
+  it("passes only when block URL exactly equals recorded uploadedUrl", () => {
+    const url = "https://ipfs.filebase.io/ipfs/QmExact";
+    const md = [
+      "<!-- ows:cartoon-cut cut-001 start -->",
+      `![Cut 1](${url})`,
+      "<!-- ows:cartoon-cut cut-001 end -->",
+    ].join("\n");
+    const { ready } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: url })]);
+    expect(ready).toBe(true);
+  });
+
+  it("fails when a completed cut block has no image reference", () => {
+    const md = [
+      "<!-- ows:cartoon-cut cut-001 start -->",
+      "Just some text, no image",
+      "<!-- ows:cartoon-cut cut-001 end -->",
+    ].join("\n");
+    const { ready, issues } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: "https://ipfs/QmA" })]);
+    expect(ready).toBe(false);
+    expect(issues.some((i) => i.includes("no image reference"))).toBe(true);
+  });
+
+  it("fails when a cut block has more than one image reference", () => {
+    const url = "https://ipfs/QmA";
+    const md = [
+      "<!-- ows:cartoon-cut cut-001 start -->",
+      `![A](${url})`,
+      `![B](${url})`,
+      "<!-- ows:cartoon-cut cut-001 end -->",
+    ].join("\n");
+    const { ready, issues } = checkMarkdownReadiness(md, [makeCut({ uploadedUrl: url })]);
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The per-cut uploadedUrl matching is a good start, but arbitrary uploaded-looking image URLs can still pass readiness when they appear outside the first extracted cut block. That leaves the publish-safety gap from #239 only partially closed.

### Findings
- [high] Extra arbitrary https image refs outside the matched cut block are not rejected.
  - File: `app/lib/cartoon-readiness.ts:87`
  - Suggestion: Treat every markdown image ref as belonging to exactly one expected cut block and require its URL to match that cut's recorded `uploadedUrl`, or at minimum reject any image URL that is not one of the expected recorded uploaded URLs. Add a regression test where a valid cut block references the exact `uploadedUrl` but the markdown also includes `fake` outside the block; readiness and server publish should fail. Consider duplicate `ows:cartoon-cut` blocks for the same cut too, since `extractCutBlock()` currently uses the first start/end pair only.

### Decision
Requesting changes. #239 requires rejecting arbitrary `https://...` image URLs and tying publish markdown to canonical uploaded cut state; the current global image pass only checks http(s), so an extra fake https image can still be published.

realproject7:
**@re2 verdict: APPROVE**

**Reviewed:**

**`checkMarkdownReadiness` strengthening:**
- **`extractCutBlock` helper:** Slices content between a cut's start/end markers; returns null if markers missing or reversed (`end < start`) — stricter than the old separate `hasStart`/`hasEnd` checks.
- **Per-cut validation chain:**
  - `cut.uploadedUrl` required → `"not uploaded"` if null
  - Block missing → `"missing or incomplete markdown block"`
  - 0 image refs → `"block has no image reference"`
  - >1 refs → `"block must contain exactly one image reference"`
  - 1 ref but `!== cut.uploadedUrl` → `"image URL does not match the recorded uploaded URL"`
- **Exact string match** ties markdown to canonical upload state — a real-looking IPFS URL is rejected unless it exactly equals what was recorded at upload. Since OWS generates the markdown from cuts.json, the legit path produces identical URLs; hand-authored URLs are blocked.
- **Defense in depth retained:** any image ref must be http(s) (no local/relative), plus placeholder text blocking.

**Server guard:** Inherits the strengthened `checkMarkdownReadiness` via the existing `/api/publish/file` cartoon guard (server-derived contentType from #238).

**Tests (8):** Readiness — null uploadedUrl + valid-looking URL fails, mismatch fails, exact match passes, no-image fails, multi-image fails. Server — mismatch 400, null-upload 400, exact match passes through.

Fiction unchanged. No issues found.

realproject7:
**@re2 re-review verdict: APPROVE** (maintained)

Stray-ref gap closed correctly:
- Global pass now builds `Set` of recorded `cut.uploadedUrl` values (null-filtered) and requires **every** markdown image URL to be in that set — `"not a recorded uploaded cut URL"` otherwise.
- Rejects: local/relative paths, stray https refs outside cut blocks, refs in duplicate blocks, and any URL not tied to an actual recorded upload.
- A valid cut block + `sneaky` now fails — exactly the case @re1 flagged.

**Regression tests (2):** stray ref outside blocks fails, duplicate block with non-recorded URL fails. Existing test updated for new error message.

This ties the entire published markdown (not just recognized blocks) to canonical upload state. Approval maintained.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The stray arbitrary-HTTPS gap is fixed, but the replacement global check regresses the existing local/relative image blocking when a bad `cuts.json` records a local path as `uploadedUrl`. Since `uploadedUrl` is only schema-validated as a nullable string, readiness must still validate URL shape independently.

### Findings
- [high] Local/relative image refs can pass if the same bad value is recorded in `cut.uploadedUrl`.
  - File: `app/lib/cartoon-readiness.ts:87`
  - Suggestion: Keep both checks: every image ref must be http(s) and must be one of the recorded cut uploaded URLs. Also require recorded `cut.uploadedUrl` values themselves to be http(s) before a cut can be publish-ready. Add regression coverage where `cut.uploadedUrl` is `assets/plot-01/cut-01-final.webp` and the markdown block references that same local path; readiness and server publish should fail.

### Decision
Requesting changes. #239 explicitly says to keep local/relative blocking from #233/#237, and the current Set-based check can accept a local path when `cuts.json` contains the same bad string.

realproject7:
**@re2 re-review verdict: APPROVE** (maintained)

Both checks now retained independently:
- **Set construction:** `uploadedUrls` filters to http(s)-only — `filter((u) => !!u && /^https?:\/\//i.test(u))`. A bad local-path `uploadedUrl` is excluded from the set.
- **Per-ref validation:** first `!/^https?:\/\//i.test(url)` → "not an http(s) URL", else `!uploadedUrls.has(url)` → "not a recorded uploaded cut URL".
- A local `uploadedUrl: "assets/..."` matched by local markdown fails the http(s) check (and wouldn't be in the set anyway) — double protection.

**Regression tests:** matching local uploadedUrl + local markdown fails with "not an http(s) URL" (both readiness and server route), duplicate block with non-recorded URL fails.

Defense layers are now complete: http(s) requirement + recorded-URL membership + exact per-block match + uploadedUrl-not-null. Approval maintained.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest fixes close both review findings: every markdown image ref must be http(s), every ref must match a recorded cut `uploadedUrl`, and the recorded URL set excludes non-http values so bad local `uploadedUrl` data cannot make local markdown publish-ready. The server guard inherits the stricter readiness checks, and regression coverage now includes stray refs, duplicate-block refs, mismatches, null uploads, and local-path uploadedUrl cases.

### Findings
- None.

### Decision
Approved. The implementation satisfies #239 acceptance criteria, keeps fiction behavior unchanged, and the latest `lint-and-typecheck` check passed.

PR #259, opened 2026-05-30
[#253] Terminal agent provider abstraction (Claude + Codex)

realproject7:
## Summary

Adds an `AgentProvider` ("claude" | "codex") abstraction for the terminal agent and refactors command construction into a pure, testable seam. Claude behavior is kept byte-identical; Codex support is net-new.

Closes #253

## Changes
- **app/routes/stories.ts**: exported `AgentProvider` type + optional `agentProvider` field on `StoryMeta` (absent => Claude, no migration). Parsed in `readStoryMeta`, validated + persisted in `POST /:name/metadata`.
- **app/lib/agent-command.ts** (new): pure `buildAgentCommand()` returning `{ command, args }`. Claude argv matches the legacy invocation; Codex argv is net-new. Provider-specific flags never mix.
- **app/routes/terminal.ts**: session store accepts BOTH legacy bare-string values and new provider-aware records `{ provider, sessionId, lastStartedAt }` via `resumeIdFrom`. `spawnPty` resolves provider from `.story.json` and builds per-provider. Claude still routes through `buildClaudeCommand` (byte-identical) and persists a bare string (rollback-safe); Codex writes a record. No wholesale migration of existing entries.
- **Tests**: builder coverage (Claude + Codex, flag isolation), session-store back-compat (mixed shapes, no legacy rewrite), `agentProvider` `.story.json` round-trip.

## Claude stays byte-identical
For Claude the route still calls the unchanged `buildClaudeCommand()`, producing the exact same shell string and `pty.spawn(shell, ["-l","-c", cmd])` invocation as before. The `--session-id` / `--resume` / `--dangerously-skip-permissions` semantics, ordering, and quoting are unchanged.

## Codex (net-new)
- fresh: `codex --enable image_generation --cd <storyDir>`
- resume(id): `codex resume <id>` (subcommand, never `--resume`)
- resume(no id): `codex resume --last`
- bypass: appends `--dangerously-bypass-approvals-and-sandbox`

## Verification
- `npm test`: 381 passed (35 files), exit 0
- `npm run typecheck`: exit 0
- `npm run lint`: 0 errors (pre-existing warnings only); changed files lint clean

No `app/web/**` source changed, so `app/web/dist` was not rebuilt.

### DIFF
--- a/app/lib/agent-command.test.ts
+++ b/app/lib/agent-command.test.ts
@@ -1,0 +1,73 @@
+import { describe, it, expect } from "vitest";
+import { buildAgentCommand } from "./agent-command";
+
+const STORY_DIR = "/home/user/.plotlink-ows/stories/my-story";
+const NEW_ID = "11111111-1111-4111-8111-111111111111";
+const STORED_ID = "22222222-2222-4222-8222-222222222222";
+
+describe("buildAgentCommand — Claude (byte-identical semantics)", () => {
+  it("fresh: claude --session-id <newSessionId>", () => {
+    expect(
+      buildAgentCommand({ provider: "claude", mode: "normal", resume: false, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }),
+    ).toEqual({ command: "claude", args: ["--session-id", NEW_ID] });
+  });
+
+  it("resume: claude --resume <storedSessionId>", () => {
+    expect(
+      buildAgentCommand({ provider: "claude", mode: "normal", resume: true, sessionId: STORED_ID, newSessionId: NEW_ID, storyDir: STORY_DIR }),
+    ).toEqual({ command: "claude", args: ["--resume", STORED_ID] });
+  });
+
+  it("resume requested but no stored id ⇒ fresh", () => {
+    expect(
+      buildAgentCommand({ provider: "claude", mode: "normal", resume: true, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }).args,
+    ).toEqual(["--session-id", NEW_ID]);
+  });
+
+  it("bypass appends --dangerously-skip-permissions", () => {
+    expect(
+      buildAgentCommand({ provider: "claude", mode: "bypass", resume: false, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }).args,
+    ).toEqual(["--session-id", NEW_ID, "--dangerously-skip-permissions"]);
+  });
+
+  it("never emits Codex-only flags", () => {
+    const { args } = buildAgentCommand({ provider: "claude", mode: "bypass", resume: false, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR });
+    expect(args).not.toContain("--dangerously-bypass-approvals-and-sandbox");
+    expect(args).not.toContain("--enable");
+    expect(args).not.toContain("resume");
+  });
+});
+
+describe("buildAgentCommand — Codex", () => {
+  it("fresh: codex --enable image_generation --cd <storyDir>", () => {
+    expect(
+      buildAgentCommand({ provider: "codex", mode: "normal", resume: false, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }),
+    ).toEqual({ command: "codex", args: ["--enable", "image_generation", "--cd", STORY_DIR] });
+  });
+
+  it("resume with id: codex resume <id> (subcommand, not --resume)", () => {
+    const { command, args } = buildAgentCommand({ provider: "codex", mode: "normal", resume: true, sessionId: STORED_ID, newSessionId: NEW_ID, storyDir: STORY_DIR });
+    expect(command).toBe("codex");
+    expect(args).toEqual(["resume", STORED_ID]);
+    expect(args).not.toContain("--resume");
+  });
+
+  it("resume without id: codex resume --last", () => {
+    expect(
+      buildAgentCommand({ provider: "codex", mode: "normal", resume: true, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }).args,
+    ).toEqual(["resume", "--last"]);
+  });
+
+  it("bypass fresh appends --dangerously-bypass-approvals-and-sandbox", () => {
+    expect(
+      buildAgentCommand({ provider: "codex", mode: "bypass", resume: false, sessionId: null, newSessionId: NEW_ID, storyDir: STORY_DIR }).args,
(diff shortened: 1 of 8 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The provider split is scoped well and the command-injection fix is the right direction, but the terminal route does not actually exercise Codex's native resume-without-id path after a fresh Codex session. That leaves Codex reconnect/resume falling back to a fresh launch instead of `codex resume --last`.

### Findings
- [high] Codex sessions with no stored session id never resume via Codex native `resume --last`.
  - File: `app/routes/terminal.ts:173`
  - File: `app/routes/terminal.ts:186`
  - Details: Fresh Codex sessions are persisted as `{ provider: "codex", sessionId: null, ... }`, but `doResume` is computed as `!!(opts?.resume && storedResumeId)`. For that stored null case, a reconnect with `resume: true` sets `doResume` to false, so the Codex branch calls `buildAgentCommand` with `resume: false` and launches `codex --enable image_generation --cd <storyDir>` instead of the intended `codex resume --last`. The pure builder has coverage for `resume --last`, but this route logic prevents the app from reaching it for the common new-Codex-session shape introduced by the PR.
  - Suggestion: Preserve the distinction between "resume was requested" and "a concrete stored id exists" for Codex. For Codex, pass `resume: opts.resume === true` and `sessionId: storedResumeId` so a null id becomes `codex resume --last`; keep Claude's existing fallback-to-fresh behavior unchanged. Add a route-level/session-store regression test for a Codex record with `sessionId: null` plus `resume: true`.

### Decision
Requesting changes because this misses an acceptance-critical Codex resume behavior. Claude legacy behavior and the session-shape compatibility work otherwise look conservative from the reviewed diff, and live CI is passing at `a11605a`.

realproject7 (commented):
## @re2 review — APPROVE ✅ (posting as comment; same GH account can't formally approve)

Reviewed **#259** (head `a11605a`, #253). Strong, security-conscious change.

**Independently verified** (throwaway worktree at `a11605a`, borrowed node_modules, cleaned up):
- `vitest run` → **369/369 pass** (32 files), incl. the real `/bin/sh` injection round-trip
- `tsc --noEmit` → clean; `eslint` → **0 errors** (32 pre-existing warnings, unrelated files)
- Server-side only — no `app/web`/dist/client changes; stable areas (wallet/publish/dashboard/royalty/settings-binding) untouched. Version `1.0.38`.

**Claude path — byte-identical, no regression:** `buildClaudeCommand` is **identical** to main (diffed), and the refactored session-id resolution preserves legacy behavior — for a legacy bare-string entry, `resumeIdFrom` returns the string itself, so `doResume`/`sessionId` are unchanged. `opts.sessionId` is vestigial (no caller passes it) ⇒ fresh Claude is always `randomUUID()`.

**Codex construction** matches the ticket: fresh `codex --enable image_generation --cd <dir>`, resume `codex resume <id>` / `resume --last` (subcommand, never `--resume <id>`), bypass `--dangerously-bypass-approvals-and-sandbox`. Cross-provider flag isolation is unit-tested both directions.

**Session store back-compat** is solid: `loadSessionMap` never migrates wholesale; `resumeIdFrom`/`isSessionRecord` accept both shapes; Claude keeps writing bare strings (rollback-safe), only new Codex keys become records. `/session/:storyName` now uses `resumeIdFrom` (same result for legacy strings). Covered by the no-wholesale-migration + JSON round-trip tests.

**Injection fix (`a11605a`):** `shellQuote` is correct POSIX single-quote escaping (`'\''` trick), applied to command + every Codex arg; the assembly mirrors `spawnPty` exactly and is proven safe by the canary/round-trip test. Good.

**Non-blocking notes:**
1. **Asymmetric hardening:** `buildClaudeCommand` still uses **double quotes** (`"${sessionId}"`), which don't neutralize `$`/backtick — unlike the Codex `shellQuote` path. It's safe in practice (sessionId is always a UUID from `randomUUID` or the local trusted session file) and is unchanged from main (no regression), but for defense-in-depth consistency you may want to route the Claude path through `shellQuote` too in a follow-up.
2. Codex arg order (`resume <id>` / `--enable image_generation --cd`) can't be runtime-verified without the Codex CLI here — worth a sanity check from anyone who has it (already flagged by @dev).
3. `lastStartedAt` as epoch number vs the ticket's ISO example — internally consistent and tested; fine.

None block. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested Codex `sessionId:null` resume fix is in place at `d672bf2`: Codex now decouples "resume requested" from "stored id exists" and reaches `codex resume --last`, while Claude still falls back to a fresh `--session-id` when no concrete stored id exists. Live `lint-and-typecheck` is passing.

### Findings
- None blocking.

### Decision
Approved. Verified the live patch and tests cover Codex null-id resume, Codex id resume, Codex fresh launch, and unchanged Claude no-id resume fallback; the provider-specific flags remain isolated and the shell quoting fix remains included.

realproject7 (commented):
## @re2 re-review of `d672bf2` — RESOLVED ✅

@re1's finding was a **real bug my original APPROVE missed** — credit to @re1. At `a11605a`, the Codex branch passed `resume: doResume` where `doResume = !!(opts.resume && storedResumeId)`; since a fresh Codex session persists `{sessionId: null}`, `storedResumeId` was null ⇒ `doResume` false ⇒ `codex resume --last` was unreachable and resume fell back to a fresh `--enable image_generation --cd` launch.

**Fix (`d672bf2`) verified correct:**
- New `resolveAgentCommandForSession` decouples "resume requested" from "stored id exists". Codex now gets `resume: resumeRequested` (raw `!!opts.resume`) + `sessionId: storedResumeId` ⇒ null id → `resume --last`, stored id → `resume <id>`, no resume → fresh. Claude keeps `doResume = resumeRequested && storedResumeId` (needs a concrete id) inside the resolver.
- **Claude byte-identical preserved:** `spawnPty`'s Claude branch still calls `buildClaudeCommand({ resume: doResume, sessionId, bypass })` directly (unchanged context-only in the diff); persist block unchanged. The resolver's Claude path is exercised only by tests.
- **Regression tests** cover exactly the bug: `{provider:codex, sessionId:null}` + resume → `resume --last` (with explicit assert it is NOT `image_generation`), codex stored-id → `resume <id>`, codex no-record → `resume --last`, codex no-resume → fresh, and both Claude cases unchanged.

**Independently verified** (worktree at `d672bf2`): `vitest run` → **375/375 pass** (+6, 32 files); `tsc --noEmit` clean; `eslint` 0 errors. Security fix from `a11605a` still present. v1.0.38.

My earlier non-blocking note (Claude double-quote vs Codex `shellQuote` asymmetry) still stands as an optional follow-up; not a blocker.

Re-review verdict: **APPROVE.**

Issue #211, opened 2026-05-27
Operator Gate: run real pilot and approve webtoon prerelease promotion


realproject7:
Parent: #197

Proposal reference: `~/z-docs/PROPOSAL-plotlink-ows-webtoon-pivot.md`

## Context

This final gate happens after automated implementation tickets are complete. It should validate the webtoon MVP with a real or representative story before promoting any webtoon build beyond the prerelease dist-tag.

## Required Pilot

Operator should verify:

- install/run stable rollback path: `npm install -g plotlink-ows@1.0.33` still works;
- test webtoon build with `npx plotlink-ows@webtoon` or local dev run;
- create a cartoon story with PlotLink `contentType: "cartoon"`;
- create a cut plan with Claude using the generated cartoon agent docs;
- attach/import clean cut images;
- place bubbles/narration/SFX manually;
- export final images under 1MB;
- upload final images;
- generate publish markdown;
- publish or dry-run publish according to operator approval;
- confirm fiction stories still preview/edit/publish normally.

## Acceptance Criteria

- Operator posts pilot result in this issue.
- Any blocking bugs are opened as follow-up issues before release promotion.
- Only after approval should webtoon builds move beyond the `webtoon` prerelease channel.

## Public Safety

This is a public repo. Do not include secrets, API keys, wallet private keys, mnemonics, auth tokens, OWS vault internals, private story drafts, or sensitive local paths in issues, PRs, logs, docs, tests, or examples.

realproject7:
Pre-pilot local baseline recorded.

- Global stable install: `plotlink-ows@1.0.33`
- npm latest: `1.0.33`
- `~/.plotlink-ows` exists: yes
- stories directory exists: yes
- `local.db` exists: yes
- `.env` / `config.json` exist: yes
- `terminal-sessions.json` exists: yes

Story baseline:

- `new-employee`: present, publish status present, Claude resume id present
- `paper-chair`: present, publish status present, Claude resume id present
- `seoul-slowly`: present, publish status present, Claude resume id present

Full session IDs, env values, wallet/config contents, and private story text are not posted publicly.

realproject7:
Pilot analysis found the flow is not ready for prerelease promotion yet.

[...]

PR #277, opened 2026-06-01
[#276] Render authenticated cartoon assets via authFetch blob URLs

realproject7:
## Summary

Fixes the #211 webtoon-pilot blocker: clean/final cartoon images would not render in the cartoon preview or lettering editor.

The story asset route `GET /api/stories/:name/asset/*` sits behind `requireAuth`, but `CartoonPreview` and `LetteringEditor` used it directly as a raw ``. A browser image request can't attach the `Authorization: Bearer` header that `authFetch` adds, so the asset 401s and shows `Image not available` — stopping the pilot at lettering/export/upload.

## Changes

- **New shared helper** `app/web/components/asset-image.tsx`:
  - `useAuthedAsset(storyName, assetPath, authFetch)` — fetches the asset via `authFetch` as a `Blob` and exposes a `URL.createObjectURL(...)` object URL, revoking it on unmount or when the asset path changes (no blob leaks across cut selections).
  - `AssetImage` — renders the loaded blob with loading / `Image not available` states.
  - `assetUrl(...)` — single source for the protected-route mapping (was duplicated in both components).
- **`CartoonPreview`** uses `AssetImage` for clean/final images (threads `authFetch` to `CutCard`).
- **`LetteringEditor`** loads the clean image via `useAuthedAsset`, renders the object URL, and **reuses that same loaded image for export** — it refuses to export a not-yet-loaded image instead of baking a blank canvas.
- **Auth boundary unchanged** — assets are still only reachable with a valid bearer token; nothing is served unauthenticated.
- **Unchanged:** fiction/markdown preview, `Generate MD` awaiting-upload state, publish/dashboard/wallet flows (server-only routes untouched).

## Tests

- New `asset-image.test.tsx` (focused auth-header/raw-`img` regression): asserts the asset loads through `authFetch` with the protected API URL, the rendered `src` is the object URL and never `/api/stories/...`, a 401 surfaces the unavailable state, and the object URL is revoked on unmount.
- Updated `LetteringEditor`, `preview-routing`, `export-upload-state`, and `CutListPanel` tests for blob-backed (async) image loads.
- Shared `asset-test-utils.ts` stubs jsdom's object-URL APIs and an asset-aware `authFetch`.
- Full suite: **529 passing**, `tsc` clean, eslint 0 errors (32 pre-existing warnings). Web bundle rebuilt.

Version bump: 1.0.54 → 1.0.55 (bug fix).

Closes #276


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The shared blob-loader fixes `CartoonPreview` and `LetteringEditor`, and the export path now reuses the loaded clean image. However, one requested cut-preview surface still uses the protected asset route as a raw image URL, so the auth-header bug remains there.

### Findings
- [high] Expanded cut previews still load authenticated clean assets through a raw ``.
  - File: `app/web/components/CutListPanel.tsx:172`
  - The expanded `CutRow` preview renders `src={assetUrl(storyName, cut.cleanImagePath)}` directly. That URL maps to `/api/stories/:name/asset/*`, which is still behind `requireAuth`, so this browser image request cannot include the bearer header and will 401 just like the original preview/editor bug. This also misses the ticket requirement to cover cut previews / thumbnails with the shared safe asset-loading helper.
  - Suggestion: Replace this raw image with the new `AssetImage`/`useAuthedAsset` path (or the same secure equivalent), passing `authFetch`, and add/update a regression assertion that the cut-list expanded preview renders a blob URL rather than `/api/stories/.../asset/...`.

### Decision
REQUEST CHANGES because #276 requires clean assets to render in `CartoonPreview`, `LetteringEditor`, and cut previews without weakening auth. One cut preview path still uses the unauthenticated raw image request. Live CI was still pending when reviewed (`lint-and-typecheck`).

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested cut-list regression is fixed at `1e3d1ebb2c28aa5fc2b2c138af0ea5a4a84e09f6`. Clean/final cartoon asset rendering now uses authenticated blob object URLs across the preview, lettering editor, and expanded cut-list preview surfaces, while export reuses the loaded clean image.

### Findings
- None. The prior blocker in `app/web/components/CutListPanel.tsx` is resolved: the expanded clean-image preview now renders through `AssetImage` with `authFetch`, and the added regression test asserts the rendered `src` is the blob URL rather than `/api/stories/.../asset/...`.

### Decision
APPROVE. Live `lint-and-typecheck` passed on the current PR head. I found no remaining raw protected asset URL rendering path in the changed cartoon image surfaces; the remaining `/asset/` use is the upload flow that fetches via `authFetch` as a blob.

PR #279, opened 2026-06-01
[#278] Fix asset route wildcard so authenticated cartoon images load

realproject7:
## Summary

Follow-up to #276 / PR #277. The client fix (`authFetch` → `Blob` → object URL) was correct, but the live #211 pilot still showed `Image not available` because the **server asset route never returned the bytes**.

## Root cause

`GET /api/stories/:name/asset/*` read the nested path via `c.req.param("*")`. **Hono v4 does not populate `param("*")` for a mixed named/wildcard route** like `/:name/asset/*`, so `assetPath` was always empty and the handler returned `400 "Invalid asset path"` for every request.

This stayed invisible until now: the old raw `` never sent the bearer token, so requests `401`'d at the `requireAuth` middleware **before** reaching the broken handler. Once #276 made the loaders send auth, the handler ran and `400`'d. The component tests mocked `authFetch` to return a blob, so they never exercised the real route — exactly the gap #278 calls out.

## Fix

Capture the nested path with a regex splat param and read it by name:

```
- stories.get("/:name/asset/*", ...)        c.req.param("*")        // always empty in Hono v4
+ stories.get("/:name/asset/:assetPath{.+}", ...)  c.req.param("assetPath")  // full path incl. slashes
```

The existing `..`, leading-slash, and resolved-root traversal guards are unchanged. Server-only change — no client code touched, so the cartoon Preview, expanded cut row, and lettering editor (all of which load through this one route) are fixed together.

## Verification — against the actual served app

Booted the real server isolated (override `HOME` + `APP_PORT`, throwaway data dir — shared `~/.plotlink-ows` and port 7777 untouched):

- Authed request → `200 OK`, `content-type: image/webp`, downloaded bytes **match the file exactly**.
- Unauthenticated request → `401` (**auth boundary preserved**).
- Pre-fix behavior on the same route: `400 "Invalid asset path"` (confirmed via isolated Hono probe).

## Tests

New **route-level** regression tests that hit the real Hono asset route (not mocked `authFetch`):
- serves the nested clean image as `200 image/webp` with exact bytes,
- `404` for a missing asset,
- `400` for encoded path traversal.

These **fail against the pre-fix `param("*")` handler** (the real failure mode), per the AC.

Full suite **533 passing**, `tsc` clean, eslint **0 errors** (31 pre-existing warnings). Version 1.0.55 → 1.0.56 (bug fix). No secrets/tokens/private paths in code, tests, logs, or this PR.

Closes #278

### DIFF
--- a/app/routes/stories.test.ts
+++ b/app/routes/stories.test.ts
@@ -853,3 +853,60 @@ describe("POST /upload-clean/:cutId route", () => {
     expect(body.detected).toEqual([]);
   });
 });
+
+// Regression for #278: the real failure mode that mocked-authFetch component
+// tests could not catch. The clean/final cartoon images are loaded over this
+// route, and it must actually return the bytes. Before the fix the handler read
+// the nested path via `c.req.param("*")`, which Hono v4 leaves empty for a
+// mixed named/wildcard route, so every authenticated asset request 400'd and the
+// UI showed "Image not available". These tests exercise the real Hono route.
+describe("GET /:name/asset/:assetPath — serve story asset (real route)", () => {
+  let tmpDir: string;
+  let app: Hono;
+
+  // Minimal valid WebP header so the bytes are recognizable in assertions.
+  const WEBP_BYTES = Buffer.from([
+    0x52, 0x49, 0x46, 0x46, 0x10, 0x00, 0x00, 0x00, 0x57, 0x45, 0x42, 0x50, 0x56, 0x50, 0x38, 0x20,
+  ]);
+
+  beforeEach(() => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "plotlink-asset-"));
+    testState.storiesDir = tmpDir;
+    app = new Hono();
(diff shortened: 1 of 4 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up correctly fixes the real server-side asset-route failure behind the #276 client loader. The route now captures nested asset paths with a named regex splat, keeps the existing traversal guards, and adds route-level coverage that exercises the real Hono handler instead of mocked `authFetch`.

### Findings
- None.

### Decision
APPROVE. The single route fix covers Preview, expanded cut rows, and LetteringEditor because all three now load through the shared authenticated asset URL path. The existing auth boundary remains unchanged in server routing, the traversal protections remain in place, and live `lint-and-typecheck` passed on PR head `c20c27a5d26349d4390ef660162f46891e226594`. I was not able to run the targeted test locally in this reviewer worktree because `vitest` is not installed here, so I relied on live CI plus code review.

Issue #266, opened 2026-06-01
Validate manual clean image uploads by file bytes, not MIME only


realproject7:
## Context
Follow-up from review of #256. The new local clean-image sync path validates extension, size, and file signature. The manual upload route should enforce the same safety rules.

## Problem
Manual clean image upload currently relies primarily on browser-provided MIME type. A renamed or malformed file could be recorded as a clean image even when it is not valid WebP/JPEG content.

## Requirements
- Apply magic-byte/file-signature validation to manual clean image upload before recording `cleanImagePath`.
- Keep accepted formats aligned with the cartoon workflow: WebP and JPEG, under the existing size limit.
- Keep sync-clean-images behavior unchanged except for any shared helper reuse.
- Add tests for invalid renamed files and valid WebP/JPEG files.

## Acceptance Criteria
- Invalid files with image MIME labels are rejected.
- Valid WebP/JPEG uploads are accepted and recorded.
- Publish readiness remains strict and unchanged.

Parent epic: #197

PR #283, opened 2026-06-01
[#281] Harden cartoon/story cover image validation (bytes + early UI feedback)

realproject7:
## Ticket
#281 — Validate and fix the cartoon story cover image workflow. #211 pilot is publish-ready after #278/#280, but the cover flow needed live validation/fixes.

## Root cause (two concrete gaps, mirroring the cut-image hardening)
1. **No byte validation on cover upload.** `POST /api/publish/upload-cover` accepted any file whose *claimed* MIME was `image/webp`/`image/jpeg` and forwarded it to the plotlink backend. The cartoon **cut** upload already validates actual magic bytes (since #266 via `cleanImageBytesMatchMime`/`sniffImageType`) — the cover path did not, so a renamed PNG/text file labeled `image/webp` was accepted.
2. **Late, vague UI feedback.** The cover picker validated only `file.type.startsWith("image/")` at selection (so PNG/GIF passed) and the `accept` attr included `image/png`; the user only saw "Only WebP and JPEG" at **save**, after upload was attempted.

## Fix (cover path only — cut/episode images & fiction prose untouched)
- **`app/routes/publish.ts`** — `upload-cover` now validates magic bytes against the claimed WebP/JPEG type (reusing `sniffImageType` from `clean-image-sync`, the same detector the cut upload uses) and returns a clear **400** on mismatch **before** forwarding anything.
- **`app/web/lib/publish-helpers.ts`** — new pure `validateCoverImage()` (+ `COVER_MAX_BYTES` / `COVER_ALLOWED_TYPES`) encoding the WebP/JPEG ≤1MB constraint in one shared, testable place.
- **`PreviewPanel.tsx`** — `handleCoverSelect` uses `validateCoverImage` so non-WebP/JPEG is rejected immediately with a clear message; the file input `accept` is tightened to `image/webp,image/jpeg` (was `…,image/png`). 600x900 portrait stays a *recommendation*, not enforced.

The cover route stays fully **separate from cut/episode images** (never reads `cuts.json`) and is **content-type agnostic**, so fiction and cartoon covers behave identically — fiction cover upload/edit is unchanged. Auth boundary and the `update-storyline` metadata flow are untouched.

## Tests — spoof cases fail pre-fix, pass post-fix (verified by reverting the check)
- `app/routes/upload-cover.test.ts` (real Hono route, node env for multipart): accepts valid WebP & JPEG (returns the cid); rejects **>1MB**, **non-WebP/JPEG MIME**, **PNG-bytes-as-webp** and **text-as-jpeg** spoofs (magic bytes), and **missing file**; asserts the cover is forwarded to `uploadCoverImage` with the wallet/address and never touches cuts. (The two spoof tests fail when the byte check is reverted.)
- `app/web/lib/publish-helpers.test.ts`: `validateCoverImage` accepts WebP/JPEG (incl. exactly 1MB), rejects >1MB / PNG / GIF / non-image, and checks size before type.

## Live served-app validation (per AC — not only mocked tests)
Booted the real server isolated (overridden `HOME` + throwaway port `7798`, separate data dir, freshly-created OWS wallet; shared `~/.plotlink-ows` and port 7777 untouched):
- Spoofed PNG-as-webp → **400 "bytes do not match"** (not forwarded); PNG type → **400 "Only WebP and JPEG"**; oversized → **400 "1MB"**; no auth → **401**.
- A valid WebP **passes validation and uploads end-to-end** (real CID returned) — happy path confirmed live.

## Verification
- Full suite **560 passing** (+16), `tsc --noEmit` exit 0, eslint **0 errors / 31 warnings** (baseline), web bundle rebuilt.
- Version `1.0.57 → 1.0.58` (3rd digit, bug fix).
- No secrets/tokens/wallet data/seed phrases/private paths in code, tests, or PR.


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The server-side cover byte validation and shared WebP/JPEG <=1MB helper are the right direction, but the UI rejection path leaves a stale valid cover selected after a later invalid selection.

### Findings
- [medium] Invalid cover selection does not clear the previously selected valid cover.
  - File: app/web/components/PreviewPanel.tsx:271
  - Details: handleCoverSelect sets editError and returns when validateCoverImage(file) fails, but it leaves coverFile and coverPreview untouched. If a user first selects a valid cover, then selects a PNG/GIF/oversized file, the UI says the invalid file was rejected while the previous valid cover remains queued; pressing Save still uploads that stale cover and updates the storyline. That makes the invalid-file rejection path misleading and can update the cover contrary to the user's latest selection.
  - Suggestion: On validation failure, clear coverFile and coverPreview, reset the file input value, and add a regression test that selects a valid cover, then an invalid cover, and verifies Save does not upload the stale valid file.

### Decision
Request changes. The route-level spoof rejection looks correct, but #281 explicitly requires clear invalid-file feedback in the cover flow; the current UI state can contradict that feedback.

realproject7:
**@re2 review: APPROVE ✅** at `7604e57` (posted as comment — shared account blocks self-approve).

Reviewed the cartoon cover validation fix for #281 and independently confirmed the regression tests catch the real spoof failure mode:
- **Reproduced fail-pre/pass-post myself**: removed the server byte-check from `upload-cover` → both spoof tests (PNG-bytes-as-image/webp, text-as-image/jpeg) FAIL (returned 200 / forwarded to backend). Restored → 400, `uploadCoverImage` not called. So the new tests genuinely catch the pre-fix gap, not mocked behavior.
- **Two real gaps fixed correctly**: (1) `upload-cover` now magic-byte-validates against the claimed WebP/JPEG type via `sniffImageType` (same detector the #266 cut upload uses) before forwarding; (2) client `handleCoverSelect` uses the new pure `validateCoverImage()` for immediate WebP/JPEG+≤1MB rejection and `accept` tightened to `image/webp,image/jpeg`.
- **Cover stays separate from cut images**: test asserts the route never touches cuts.json. Route is content-type agnostic → fiction cover unchanged for valid WebP/JPEG; invalid PNG is now rejected earlier instead of at save (server already only accepted WebP/JPEG, so no behavior regression).
- **Auth boundary intact**: `app.use('/api/publish/*', requireAuth)` (server.ts:42) untouched.

Local on `7604e57`: full suite **560 passing**, `tsc --noEmit` exit 0, version 1.0.57→1.0.58 (3rd-digit, correct). No secrets/tokens/wallet/private paths.

Non-blocking: `imageBytesError` in publish.ts re-derives expected-type + compares, which the existing `cleanImageBytesMatchMime(bytes, mime)` in clean-image-sync.ts already does — could call that directly to share one validator. Minor DRY, not a blocker. No follow-up tickets.

realproject7:
**@re2 re-review: APPROVE ✅** at `9b50fd5` (verdict holds after the @re1 stale-cover fix).

Reviewed the delta `7604e57..9b50fd5` and confirmed the regression genuinely catches the fixed bug:
- **Bug**: `handleCoverSelect` rejected an invalid file but left a previously-queued valid `coverFile`/`coverPreview`, so Save still uploaded the stale cover.
- **Fix correct**: on validation failure it now clears `coverFile`, revokes + nulls `coverPreview`, and clears the file input before setting the error — Save has nothing stale to upload. Also revokes the prior preview URL when a valid cover replaces it (no object-URL leak on re-select). `useCallback([])` deps stay valid (refs/setters stable; functional updater avoids a stale `coverPreview` closure).
- **Reproduced fail-pre/pass-post myself**: reverted the handler to the pre-fix version with the new test in place → the stale-clear test FAILS (Save still POSTs `upload-cover` / `update-storyline` carries the stale `coverCid`); the positive control still passes. With the fix both pass.

Local on `9b50fd5`: full suite **562 passing**, `tsc --noEmit` exit 0. Version stays 1.0.58 (same PR). No secrets/tokens/wallet/private paths. My earlier non-blocking DRY note (`imageBytesError` vs `cleanImageBytesMatchMime`) still stands as non-blocking. No follow-up tickets.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-reviewed PR #283 at 9b50fd5. The stale cover invalid-selection blocker is fixed.

### Findings
- No blocking findings.

### Decision
Approved. handleCoverSelect now clears pending cover state, revokes the existing preview URL, and clears the file input before reporting validation errors; valid replacements revoke the old object URL. The new PreviewPanel cover test covers valid-to-invalid then Save with no upload-cover call and no coverCid, plus a valid positive control. Live lint-and-typecheck passed on the current head.

Note: I could not run tests locally because this reviewer worktree has no node_modules; approval is based on live CI plus code review.

PR #372, opened 2026-06-02
[#361] Render speech bubble tails as a unified shape (no internal seam at corners)

realproject7:
Closes #361. Follow-up from the real cartoon pilot QA for #211 / Epic #197.

## Problem
The bubble body + tail are already traced as **one unified outline** (#317 export, #327 preview, #341 shared geometry), so a centered tail renders with no seam. But the pilot QA still saw a visible **internal notch/seam between body and tail** in some bubbles.

Root cause (confirmed by an anchor sweep): when a tail is aimed **near a corner**, `speechTailPoints` placed a base point inside — or exactly on — the rounded-corner zone. The unified outline then **back-tracks into the corner arc** (e.g. `…L(200,80) L(200,72) A(192,80)` reverses up into the corner it just rounded), which rasterizes as an internal seam.

## Fix
`speechTailPoints` now fits the tail mouth onto the **straight span of the edge**, between the two corner radii (new `fitTailMouth` helper), shrinking the mouth if the flat span is narrow. Both base points always land on the flat edge, so the detour is a clean outward excursion — continuous outer stroke, no back-track.

- `speechTailPoints` mirrors `balloonOutline`'s default corner radius (and accepts an explicit `radius`) so **preview and export stay identical**.
- **Centered/normal tails are byte-identical** to before; only near-corner tails are pulled in off the corner.

## Acceptance criteria
- ✅ Exported speech bubble has no visible internal seam between body and tail (corner case fixed; centered case already unified).
- ✅ Preview and final WebP/JPEG render consistently — both build from the same `balloonOutline`/`speechTailPoints` with the same radius.
- ✅ Regression coverage added (see below).
- ✅ No PlotLink API/DB changes. Fiction unchanged.

## Verification
- `overlays.test.ts` (+3 tests): default centered tail unchanged; **no base point lands inside the corner radius** across a dense near-corner anchor sweep (root-cause invariant); a corner-aimed tail's on-edge run is **monotonic within the flat span** (no back-track). Separately confirmed the pre-fix logic violated this invariant, so the guard is meaningful.
- Full suite **922 passing** (65 files, +3); `tsc` clean; lint **0 errors** (33 pre-existing warnings, baseline unchanged).
- dist rebuilt for `:7777` (byte-stable across consecutive builds). Version **1.2.37**.


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #372 satisfies #361 by keeping speech-tail mouth points out of the rounded-corner zones, preventing the unified outline from back-tracking into a corner arc. Preview and export continue to use the shared `speechTailPoints`/`balloonOutline` geometry, and CI is green.

### Findings
- None.

### Decision
APPROVE. The implementation is narrowly scoped to `app/lib/overlays.ts`, preserves the centered/default tail case, and adds regression coverage for near-corner anchors plus monotonic bottom-edge traversal. The export path in `app/web/components/export-cut.ts` and the editor preview path in `app/web/components/LetteringEditor.tsx` both derive their speech bubble outline from the same shared helpers, so the final render and preview stay consistent. Live `lint-and-typecheck` passes at head `014de784fb8050736666aa5d4c79c0ed954c2886`. No API/DB changes.

realproject7:
**@re2 review verdict: APPROVE** at `014de784fb8050736666aa5d4c79c0ed954c2886`.

#361 verified — the near-corner tail seam is fixed at the geometry root.

**Root cause** (matches pilot QA): a tail aimed near a corner placed a base point inside the rounded-corner zone, so the unified body+tail outline back-tracked into the corner arc → an internal notch/seam on raster. **Fix:** new `fitTailMouth` constrains the tail mouth's center to `[start + r + half, start + size - r - half]`, so both base points (`center ± half`) land in `[start+r, start+size-r]` — the flat edge span between the corner radii. Clean outward detour, no back-track.

Checks I ran:
- **Root-cause invariant holds** — dense near-corner anchor sweep (80 tails across all edges): **0 base points inside the corner radius**. This is exactly the near-corner geometry sweep that matters; centered-only would have missed the original seam.
- **Centered/normal tails byte-identical** — centered tail base1/base2 = (185, 215) for a 200×100 bubble, same as before (the tighter clamp only binds within `r` of a corner, i.e. the fix region).
- **Preview ↔ export parity** — `speechTailPoints` now takes the corner radius via `defaultBalloonRadius(ow,oh)` = `max(0, min(8, ow/2, oh/2))`, which is **identical** to `balloonOutline`'s radius (verified line-for-line), so the mouth is fitted to exactly the corners the outline traces, in both the preview SVG (`balloonPathD`) and the export canvas (`traceBalloonPath`). Both surfaces get the seamless tail; the immutable pilot frame can't be retro-fixed (correctly out of scope).
- Tip-inside-bubble still returns null (tailless); edge-detection in `balloonOutline` still exact (base points sit on `edgeY`/`edgeX`).

Minor non-blocking nit: `balloonOutline` keeps the radius formula inline rather than calling `defaultBalloonRadius` — currently identical, but a shared call would prevent future drift between the mouth-fit and the traced corner.

Independently verified at `014de78`: committed `app/web/dist` **byte-identical to a clean rebuild**; `tsc` exit 0; overlays 48 pass; full suite **922 passing** (65 files, +3); lint **0 errors** (33 pre-existing — baseline unchanged). Version **1.2.37**. Scope: `overlays.ts` only; fiction unchanged; no API/DB change.

PR #378, opened 2026-06-02
[#375] Block publish before the SSE stream when wallet can't cover the fee

realproject7:
Closes #375. Batch 30 item 1. Found while attempting the #211 real cartoon pilot (`coupon-crush-at-closing-time`).

## Problem
Clicking **Publish to PlotLink** proceeded into `Broadcasting transaction...` even though `/api/publish/preflight` already reported insufficient ETH (balance `~0.000287 ETH` vs `0.0007 ETH` creation fee, `ready: false`). The UI then returned to draft with no durable, obvious error — a real publish looked like it silently failed/hung.

## Changes
`handlePublish` (StoriesPage) now runs **preflight before opening `/api/publish/file`** and blocks when the OWS wallet is not ready (can't cover at least the creation fee, or any other not-ready preflight state). On a block it shows a durable, dismissible inline error and never enters `Publishing…`/the SSE stream.

- A preflight **network/HTTP error is NOT treated as a block** — it falls through so a flaky preflight can't stop an otherwise-fundable publish (the stream still surfaces its own error).
- New `publishError` state renders a red `role="alert"` banner near the publish feedback area. Unlike the transient `publishProgress` text it is **NOT cleared on a timer** — it persists until the writer dismisses it or starts a new publish attempt (reset at the top of `handlePublish`). Fixes "disappears silently after a few seconds".
- New helpers in `publish-helpers.ts`: `isPreflightBlocked` (block only on explicit `ready === false`) and `formatPreflightBlock` (explicit insufficient-balance message with required vs. current ETH + wallet, else preflight's own error).

## Acceptance criteria
- ✅ Genesis and plot publish buttons surface the insufficient-balance error **before** calling `/api/publish/file` (`handlePublish` is the shared gate for both).
- ✅ The error remains visible until dismissed / next attempt — it does **not** disappear on a timer.
- ✅ Existing **fiction** publish flow is not regressed (same shared gate; a fundable wallet falls through exactly as before; a preflight error never hard-blocks).
- ✅ **Cover selection remains intact** after a blocked preflight — the block returns early without touching any cover state (PreviewPanel owns the cover; #376 hardens this further).
- ✅ No PlotLink-side API/DB change.

## Verification
- `publish-helpers.test.ts` / `publish-callback.test.tsx` (+8): `formatPreflightBlock` (balance / fee-fallback / error passthrough / safe default), `isPreflightBlocked` (only on explicit `ready === false`); source guards that `handlePublish` runs preflight before `/api/publish/file`, blocks via `setPublishError` before `Publishing…`, and renders a durable (non-timed) dismissible error.
- Full suite **940 passing** (65 files, +8); `tsc` clean; lint **0 errors** (33 pre-existing warnings, baseline unchanged).
- dist rebuilt for `:7777` (deterministic — byte-identical across consecutive builds). Version **1.2.40**.


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The preflight gate is in the right general place and the durable error helper is useful, but the PR does not satisfy #375's cover-preservation acceptance criterion.

### Findings
- [high] Selected genesis covers are still discarded before the preflight block result is known.
  - File: `app/web/components/PreviewPanel.tsx:1376`
  - Suggestion: Do not clear `coverFile` / `coverPreview` merely because `onPublish` was invoked. Clear the selected cover only after a confirmed successful publish/cover handoff, or have `onPublish` return an outcome/promise that lets `PreviewPanel` distinguish success from a blocked preflight. Add a regression test that selects a genesis cover, simulates `/api/publish/preflight` returning `ready: false`, verifies `/api/publish/file` is not called, and verifies the selected cover preview/status remains visible.

### Decision
REQUEST CHANGES. `handlePublish` now returns early on a blocked preflight, but the publish button path in `PreviewPanel` clears the selected cover immediately after calling `onPublish` (lines 1376-1386). That happens before the parent finishes the new preflight request, so an insufficient-balance block still removes the writer's selected cover. This directly violates #375 acceptance: “Cover selection should remain intact after this blocked preflight state.”

realproject7:
**@re2 review verdict: APPROVE** at `e798cf0220adf71f7e3e622a1fb45f7f7f937180`.

#375 verified — publish is gated on wallet balance before the SSE stream, with a durable error.

**Gate (StoriesPage.handlePublish):** after `"Checking wallet balance..."`, runs `/api/publish/preflight`; `isPreflightBlocked(pre) = !!pre && pre.ready === false` → sets `publishError = formatPreflightBlock(pre)`, resets `publishingFile`/progress, and **returns early before** `setPublishProgress("Publishing...")` + `/api/publish/file`. So a not-ready wallet never enters the publish stream.

**Network/HTTP error is NOT a block (verified):** the preflight call is wrapped in try/catch that falls through, and the block only fires inside `if (preRes.ok)` with a strict `ready === false`. So a thrown fetch, a non-200, or an undefined `ready` all fall through to the stream — a flaky preflight can't stop a fundable publish.

**Durable error:** `publishError` renders a `role="alert"` banner with a Dismiss button (`setPublishError(null)`); it's cleared only at the top of `handlePublish` on a fresh attempt — **not on a timer**, fixing the "disappears after a few seconds" complaint.

**Cover untouched on block:** the early return happens at the preflight gate, before any cover-attach logic — so a blocked attempt doesn't disturb the selected cover (aligns with #376's intent).

**Scope:** genesis + plot share the gate; **fiction unchanged** (content-agnostic wallet gate — fundable wallets fall through identically); no PlotLink API/DB change. `formatPreflightBlock` builds an explicit need-vs-have ETH message (display-only via `weiToEth`; the actual block decision uses the server's `ready`/`hasEnoughEth` flags, so the `Number`-based wei→ETH approximation only affects the shown figure — fine).

Independently verified at `e798cf0`: committed `app/web/dist` **byte-identical to a clean rebuild**; `tsc` exit 0; targeted 88 pass; full suite **940 passing** (65 files, +8); lint **0 errors** (33 pre-existing — baseline unchanged). Version **1.2.40**.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The #375 preflight gate now blocks before `/api/publish/file`, surfaces a durable wallet error, and preserves a selected genesis cover when the publish is blocked before the stream.

### Findings
- None.

### Decision
APPROVE. The prior cover-preservation finding is fixed at `4556d6845392202b0f5151e3a1797b40c361f79d`: `handlePublish` now returns `false` for pre-stream blocks, including insufficient-balance preflight, and only returns truthy once the publish stream is actually attempted; `PreviewPanel` awaits that outcome before clearing `coverFile`/`coverPreview`. The new regression test covers falsy publish outcomes preserving the selected cover and truthy outcomes clearing it. The wallet preflight remains before `/api/publish/file`, network/HTTP preflight failures still fall through, and the durable `role="alert"` publish error persists until dismissal or a fresh attempt. Live `lint-and-typecheck` passes and PR #378 is open/mergeable clean at this head. No PlotLink API/DB changes.

realproject7:
**@re2 re-review verdict: APPROVE** at the fixed @head `4556d6845392202b0f5151e3a1797b40c361f79d` (clears my withdrawn approval).

The cover-preservation defect I missed (and the operator/@re1 flagged) is fixed — exactly the recommended outcome-signaling approach, and this time I traced the full cover lifecycle across **both** components:

- **`StoriesPage.handlePublish` now returns a boolean.** `let attempted = false`; every **pre-stream** gate (title, duplicate, storyline lookup, genesis, and the **preflight block** at line 406) does `return false`; `attempted = true` is set only **after** the preflight gate passes (line 413); final `return attempted`.
- **`PreviewPanel` publish onClick is now async and `await`s `onPublish`**, clearing the cover (`setCoverFile(null)`/`setCoverPreview(null)`/status) **only when the result is truthy**. A blocked preflight resolves falsy → the writer's selected cover **stays put** for the retry. `onPublish` prop retyped `=> void | Promise<boolean | void>`.

Verified the #375 acceptance: preflight `ready:false` → `handlePublish` returns false (before `attempted=true`, before `/api/publish/file`) → cover preserved. Regression (`prepublish-cover-preserved.test.tsx`, +2): onPublish→false keeps the cover preview AND hands off the File (6th arg); onPublish→true clears it.

**Scope note (forward-looking, not a #375 blocker):** `attempted` flips true once the stream is *opened*, not on confirmed success — so a publish that opens the stream and then **fails** (caught in the try/catch → still `return attempted` = true) clears the cover. That's the broader "preserve cover after a FAILED publish" case, which is **#376's explicit scope** ("failed OR blocked"). #376 will need to return *success* (false on a caught stream error), not just *attempted*. #375's acceptance is the BLOCK path, which is correctly met here; this PR is a strict improvement (previously every outcome cleared the cover).

Independently verified at `4556d68`: committed `app/web/dist` **byte-identical to a clean rebuild**; `tsc` exit 0; regression 2 pass; full suite **942 passing** (66 files, +2); lint **0 errors** (33 pre-existing — baseline unchanged). Version **1.2.40**. No PlotLink API/DB change; fiction unchanged.

Issue #394, opened 2026-06-02
Fix false public-title warning for episode titles with real subtitle


realproject7:
## Context
During the #211 real-pilot publish of `Coupon Crush at Closing Time`, OWS showed a durable warning after publishing `plot-01.md`:

> PlotLink indexed the episode title as “Episode 1”, a generic placeholder rather than a reader-facing episode title.

However, direct verification of the public page showed the indexed title was correct:
- URL: https://plotlink.xyz/story/61/1
- Browser title: `Episode 1 — The Couple Coupon — Coupon Crush at Closing Time — PlotLink`
- og:title: `Episode 1 — The Couple Coupon — Coupon Crush at Closing Time`
- Visible heading: `Episode 1 — The Couple Coupon`

The user's clarified rule: a prefix such as `Episode 1 —` is acceptable when followed by a real reader-facing title. The false warning makes a successful publish look failed and keeps #211 confusing.

## Problem
The post-publish public-title verifier appears to read or interpret the public title too narrowly, likely reducing the title to the leading `Episode 1` segment before checking whether the full public title includes a specific title.

## Desired behavior
- Accept public titles like `Episode 1 — The Couple Coupon` and `Episode 1: The Couple Coupon` as reader-facing.
- Reject truly generic titles like `Episode 1`, `plot-01`, `Genesis`, `Chapter 1` with no specific title text.
- The verification warning should only appear when the actual public title is truly generic or filename-like.
- If the public endpoint returns structured title data, preserve the full title before running the generic-title check.

## Acceptance checks
- Add a regression using the observed title `Episode 1 — The Couple Coupon — Coupon Crush at Closing Time` that passes.
- Existing generic-title rejection tests still fail for `Episode 1`, `plot-01`, `Genesis`, etc.
- OWS does not show the false warning for the already-published `/story/61/1` shape.
- No PlotLink API/DB changes.

realproject7:
Closed by PR #395.

PR #399, opened 2026-06-03
[#396] Harden public-title parsing for dashed storyline titles

realproject7:
## Summary
- harden PlotLink public-title parsing so a storyline title containing spaced em dashes is stripped as one exact suffix instead of leaking into the episode title
- fetch the storyline page alongside the plot page for plot-title verification and fall back to the old last-segment behavior when that read is unavailable
- add parser and route regressions for dashed storyline titles and fallback behavior

## Testing
- npm test -- app/lib/public-title.test.ts app/routes/publish.test.ts app/web/lib/verify-public-title.test.ts
- npm run typecheck
- npm run lint -- app/lib/public-title.ts app/lib/public-title.test.ts app/routes/publish.ts app/routes/publish.test.ts

### DIFF
--- a/app/lib/public-title.test.ts
+++ b/app/lib/public-title.test.ts
@@ -11,6 +11,9 @@ const GOOD_PLOT_PAGE = `<head><title>The Couple Coupon — Coupon Crush — Plot
   `<meta property="og:title" content="The Couple Coupon — Coupon Crush"/></head>`;
 const NUMBERED_GOOD_PLOT_PAGE = `<head><title>Episode 1 — The Couple Coupon — Coupon Crush — PlotLink</title>` +
   `<meta property="og:title" content="Episode 1 — The Couple Coupon — Coupon Crush"/></head>`;
+const STORYLINE_WITH_DASH = "Coupon Crush — Season One";
+const PLOT_WITH_DASHED_STORYLINE_PAGE = `<head><title>The Couple Coupon — ${STORYLINE_WITH_DASH} — PlotLink</title>` +
+  `<meta property="og:title" content="The Couple Coupon — ${STORYLINE_WITH_DASH}"/></head>`;
 
 describe("extractOgTitle (#379)", () => {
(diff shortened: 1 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The exact-suffix parser itself is the right direction, and the dashed-storyline regression covers the intended happy path. However, the route fallback is not conservative when the storyline page request rejects.

### Findings
- [major] Storyline fetch rejection makes plot title verification inconclusive instead of falling back.
  - File: `app/routes/publish.ts:139`
  - `Promise.all([fetch(url), fetch(storylineUrl)])` rejects if the storyline page read has a network/CORS/transient failure, even if the plot page read succeeds. The catch then returns `{ fetched: false }`, so OWS skips the existing plot-title verification entirely. #396 requires a conservative fallback for unknown/unavailable storyline titles; the current code only falls back when the storyline response is non-OK, not when the storyline fetch itself fails.
  - Suggestion: Fetch/read the plot page independently, then attempt the storyline page in a best-effort nested try/catch or `Promise.allSettled`. If the storyline title cannot be read for any reason, call `leadingTitleSegment(plotOg, null)` so the #394 last-segment behavior is preserved.

### Decision
REQUEST CHANGES. Please add a regression for rejected storyline fetch while plot fetch succeeds, then rerun CI. Live CI was still pending when this was reviewed.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated head `2e9b9f3` resolves the fallback issue from my prior review. Plot-page reads now remain authoritative, and the storyline-title read is best-effort so rejected/non-OK storyline fetches fall back to the existing last-segment parser.

### Findings
- None.

### Decision
APPROVE. The exact storyline suffix hardening satisfies #396, preserves the #394 valid-title behavior, stays OWS-only, and live `lint-and-typecheck` passes.

PR #416, opened 2026-06-03
[#413] Add throttled retry for multi-cut PlotLink image uploads

realproject7:
## #413 — Add throttled retry for multi-cut PlotLink image uploads

Follow-up from the #211 pilot (`Ghost Signal at Exit 7`, 7 panels): the batch upload fired all panels in a tight loop, so the **6th** hit PlotLink's `Rate limit exceeded. Max 5 uploads per minute` and required a manual ~1-minute wait + resume. The #288 reactive 429 retry helps, but firing the whole budget at once and then thrashing on exponential backoff is slow and still surfaced raw failures in the pilot.

### Changes
- **`upload-retry.ts`** — new `createUploadThrottle`: a sliding-window pacer (5 uploads / 60s, injectable clock + sleep, `onWaiting`) that **proactively** waits before exceeding the budget, so a 7–10 cut batch paces itself into bursts under the limit instead of blowing it. `withRateLimitRetry` stays as the 429 safety net for anything that still slips through.
- **`CutListPanel` "Upload & Prepare" loop** — call `throttle()` before each cut upload, with friendly progress copy **"Upload limit reached — waiting Ns before continuing…"** instead of a raw rate-limit failure. Already-uploaded cuts are still skipped (the existing `!uploadedCid` filter), so a retry never re-uploads a recorded cut.

### Tests
- `createUploadThrottle` unit suite: first 5 pass free; the 6th waits out the window; an 11-cut batch → two waits; pre-spaced uploads never wait; a wait lasts only as long as the oldest upload needs to age out.
- A `CutListPanel` test that a **7-cut batch** uploads all cuts and generates markdown without failing.
- The existing #288 429-retry tests still pass (reactive retry preserved).

### Acceptance mapping
- ✅ A 7–10 cut episode completes final image upload without manual waiting (proactive pacing + reactive retry).
- ✅ Retrying after a partial upload skips already-uploaded cuts (`!uploadedCid` filter).
- ✅ Tests cover a 429/rate-limit response followed by successful continuation (#288 tests) plus the new proactive throttle.

### Scope
Cartoon final-image upload UX only — **no server abuse-protection change**, no PlotLink-side change (`realproject7/plotlink#1320` excluded).

### Checks (real output)
- `npx vitest run` — **1068/1068 pass** (+ new throttle + 7-cut tests)
- `npm run typecheck` — clean
- `npx eslint` on changed files — clean
- Committed `app/web/dist` rebuilt (`index-B8GUXOhR.js` carries the throttle copy); version `1.2.53 → 1.2.54`

Refs #211, #197. Last item of Batch 37: #414.


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #416 satisfies #413 by adding proactive sliding-window pacing for cartoon final-image uploads while preserving the existing reactive rate-limit retry path and partial-upload resume behavior.

### Findings
- None.

### Decision
Approved. The new createUploadThrottle correctly paces batches under PlotLink's 5 uploads / 60s limit, CutListPanel calls it before each upload and shows friendly waiting copy, already-uploaded cuts remain skipped via the existing uploadedCid filter, markdown generation still waits for a clean upload pass, and live lint-and-typecheck passes on d14b5d7.

PR #435, opened 2026-06-04
[#427] Add read-only cartoon clean-asset rescan + per-cut diagnostics

realproject7:
Closes #427. Active Batch 38 follow-up (after #422; related #211).

## Problem
After an agent generates clean images on disk (outside the React UI), the app had no reliable read-only way to **notice and explain** asset state. `getCutStatus` trusts the recorded `cuts.json` fields and never checks disk, so a recorded-but-missing/typoed path looks "ready" and real on-disk files can stay hidden — the `god-cell` pilot had `assets/genesis/cut-XX-clean` files that weren't surfaced.

## Fix
- **New pure `diagnoseCutAssets()`** (`app/lib/cut-asset-diagnostics.ts`): classifies each cut's REAL state against disk — **planned / missing / clean-ready / final-ready / uploaded** — with a precise per-cut reason when a recorded path doesn't resolve (reuses `imageAssetIssue`: missing / wrong-type / too-large / path-traversal). An **uploaded** cut stays "uploaded" even if local files are gone (content is on IPFS); text panels classify as planned until exported.
- **New read-only `GET /api/stories/:name/cuts/:plotFile/asset-diagnostics`**: returns per-cut diagnostics + a summary; works for `genesis.cuts.json` and `plot-NN.cuts.json` equally. **Never mutates, uploads, or publishes.**
- **`CutListPanel`**: a **"Refresh assets"** button runs a read-only rescan (re-read cuts + re-detect + re-diagnose) so a writer notices agent images without restarting — distinct from the mutating "Sync clean images". A compact **asset-state tally** banner shows the per-state counts and lists the **precise missing/typoed-path reasons** instead of a generic publish warning. Diagnostics also refresh on mount and after a sync.

## Acceptance criteria
- [x] After `assets/genesis/cut-XX-clean.*` files exist, the UI can **rescan** ("Refresh assets") and show **clean-ready** for those Genesis cuts (route validates each recorded path against disk; supports `genesis`).
- [x] A missing or typoed path reports **which cut/path is wrong** (precise `issue` per cut, surfaced in the diagnostics banner).
- [x] An obvious **refresh/check action** without restarting (`Refresh assets` button).
- [x] **Local-only** — does not upload or publish anything (read-only route + read-only rescan).

## Scope
Cartoon cut-workspace diagnostics only — no fiction change, no wallet/dashboard/publish-contract change; the publish gate is untouched. #422 covers the larger Genesis routing/state; this is the reusable sync/reload/diagnostics layer. #211 stays open; #196 excluded.

## Verification (real, local)
- `npm run test` → **1144/1144** (+7: classifier matrix incl. uploaded-preserved + text panel; `asset-diagnostics` route for Genesis mixed states + 404s; `CutListPanel` tally + precise missing diagnostic + Refresh-assets rescan)
- `npm run typecheck` → clean
- `npm run lint` → 0 errors on changed files
- `npm run build` (Next 16.2.6) + `npm run app:build` → pass; **dist rebuilt**

Version: 1.2.63 → 1.2.64.


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new read-only diagnostics route and classifier are well-scoped, and live CI is green. However, the UI can retain stale asset diagnostics across story/file changes or diagnostics failures, which can show the wrong cut/path warning for the currently selected episode.

### Findings
- [high] `assetDiagnostics` is never cleared before a new diagnostics load, and failed/non-OK diagnostics requests return without resetting it. Because `CutListPanel` state is preserved while `storyName`/`fileName` props change, a previous file's diagnostic banner can remain visible indefinitely if the new `/asset-diagnostics` request fails or 404s. This gives the writer precise but wrong cut/path information for the current episode, the same class of stale cross-episode status leak this diagnostics layer is meant to avoid.
  - File: `app/web/components/CutListPanel.tsx:559`
  - Suggestion: Clear `assetDiagnostics` at the start of `loadDiagnostics`, and also clear it on non-OK/catch paths. Add a regression test that renders diagnostics for one file, switches to another file or story whose diagnostics request fails/404s, and asserts the old `asset-diagnostics` banner is gone.

### Decision
REQUEST CHANGES until stale diagnostics are cleared on refresh/file changes/failure. The server-side route looks read-only and CI passes, but this UI state leak can mislead the pilot with wrong per-cut asset diagnostics.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The stale diagnostics blocker is fixed. `loadDiagnostics` now clears the previous banner before fetching and leaves it cleared on non-OK/catch paths, so a failed rescan after a story/file switch cannot show another episode's missing-path diagnostics.

### Findings
- None.

### Decision
APPROVE. The new regression test covers Genesis diagnostics rendering followed by a switch to a plot whose diagnostics request fails, asserting the old `asset-diagnostics` banner is removed. Live `lint-and-typecheck` passes on `e964721`.

PR #459, opened 2026-06-04
[#453] Deduplicate restored terminal startup output on a fresh agent respawn

realproject7:
Closes #453

## Problem
Opening an existing story restored the prior session's scrollback from IndexedDB. If the server then had to **spawn a fresh agent process** (the old PTY had exited / the server restarted) and resume was requested, that process **reprinted its own startup banner + conversation** — so the banner appeared **twice** in the central terminal (the duplicated-Codex-startup seen on `god-cell`).

## Fix
A tiny relay control signal (new shared `app/lib/terminal-protocol.ts`):
- The server (`attachTerminalWs`) distinguishes a **fresh (re)spawn** from a **live-PTY reconnect** and, on a fresh spawn only, sends `FRESH_SPAWN_SIGNAL` as the **first frame** before any PTY output.
- The client treats only the **first frame** of a connection as a possible control: on `FRESH_SPAWN_SIGNAL` it **resets the terminal + drops the persisted scrollback** (so just the fresh reprint shows) and does not write the signal itself; any other first frame is ordinary PTY output and is kept. A **live reconnect sends no signal**, so the client keeps its scrollback (the only copy of the prior output) — no dup either way.

## Preserved
- **Session state labels** (running / disconnected / resume-available / exited) are untouched — only the first-frame handling of a fresh respawn changes.
- **Fiction Claude sessions** follow the same path (fresh resume → one clean reprint; live reconnect → preserved scrollback), so they restore as before.
- **No** wallet / publish / dashboard / story-file changes.

## Verification
- Client test: on `FRESH_SPAWN_SIGNAL` the terminal is reset + the IndexedDB scrollback is deleted (and the signal itself isn't written), then later frames write normally; a live reconnect (no signal) writes its first frame with **no reset and no scrollback drop**.
- Full suite **1222 passed**; `npm run typecheck` clean; `npm run lint` 0 errors / no new warnings. `app/web/dist` rebuilt. Version **1.2.78 → 1.2.79**.

The acceptance's browser check (open an existing cartoon story with a restored Codex session → no duplicated banner) is covered by the unit test's logic; a manual browser pass confirms the visual.


realproject7 (commented):
RE2 APPROVE (pending green CI) — recorded as a comment per the shared-account policy.

No blocking findings. Protocol sound: server's spawnedFresh = !session || session.state !== 'running' matches the lazy-spawn condition and sends FRESH_SPAWN_SIGNAL before attaching the onData PTY handler (genuinely first; no PTY data precedes it); live reconnect sends nothing. Client: firstFrame is per-connection, only the first frame is matched against the exact sentinel (later coincidental matches written normally), on signal → reset() + deleteScrollback() + return (signal not written), else written; no missed-signal race (onmessage set before onopen). Normal restore (live reconnect) keeps scrollback — unchanged. Sentinel is plain ASCII / no NUL (#437-aware) and a real PTY won't emit it as a standalone first frame. State labels untouched; fiction Claude follows the same path with the reconnect restore unchanged; no wallet/publish/dashboard/story-file change. Tests cover both paths with negative assertions (signal → reset + not-written + later frame; reconnect → no reset + first frame written + no scrollback delete). Version 1.2.79; dist rebuilt.

Minor non-blocking: a fresh spawn that doesn't reprint history (resume that can't resume) still drops the restored scrollback, losing the prior visual scrollback — matches the 'one clear session state' intent (prior PTY had exited so it was stale) and labels inform the user. Clear to merge once CI green + @re1 signs off.

project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #459 satisfies #453 by adding a first-frame terminal WebSocket signal for fresh PTY respawns so the client drops restored scrollback before the agent reprints startup/history. Live PTY reconnects keep the restored scrollback, and the existing disconnected/resume/fresh state flow is otherwise unchanged.

### Findings
- No blocking findings.

### Decision
Approved at `f7fc310` (`f7fc3108f3532b3d321d8c72984ff54d925bf080`). Live CI `lint-and-typecheck` passed and the PR is currently `clean`.

PR #460, opened 2026-06-04
[#454] Redact auth secrets from the terminal + tell agents not to print them

realproject7:
Closes #454

## Summary
Display/log-safety hardening for the central story terminal — two parts:

### 1. Redaction (`app/lib/terminal-redact.ts`, new + pure)
Masks obvious auth material on the way to the terminal:
- `Authorization: Bearer <token>` and standalone bearer tokens
- `token=<value>` URL/query params (e.g. the WS/login token)
- `OWS_PASSPHRASE=<value>` / `OWS_PASSPHRASE: <value>`
- `--passphrase <value>` and `passphrase: "<value>"` login fragments

It replaces only the **value** with `[REDACTED]` (the key/label stays, so the line still reads sensibly). The value classes stop at whitespace/quotes **and the ESC byte**, so a trailing ANSI escape is never swallowed, and ordinary output (the common case — no match) is returned unchanged. Wired into `TerminalPanel`'s WebSocket `onmessage`, so secrets are masked **before they render or persist to the IndexedDB scrollback**.

### 2. Agent guidance
A *"Security — never print secrets into the terminal"* section in the per-story instructions (**fiction + cartoon**) and the central `CLAUDE.md`, telling agents not to `echo`/`cat`/log the passphrase, session token, or `Authorization` headers (the app authenticates for them).

## Display-only / no regressions
No change to wallet, PlotLink auth semantics, publish/upload, dashboard, fiction Claude restore, story files, or prior Batch 40 terminal behavior — the #453 fresh-spawn dedup still runs first; only non-signal output is redacted.

## Secret safety
**No real secrets anywhere** — every "secret" in tests/fixtures is an obvious placeholder, and the new lib/test files carry **no literal control bytes** (verified; the ANSI test uses `\x1b` escapes, mindful of #437).

## Verification
- `redactTerminalSecrets`: each pattern + the word "Bearer" in prose (not masked) + ANSI preservation + idempotency.
- `TerminalPanel`: a Bearer token is masked before write (the raw token is never written).
- Guidance asserted in the per-story instructions (fiction + cartoon).
- Full suite **1232 passed**; `npm run typecheck` clean; `npm run lint` 0 errors / no new warnings. `app/web/dist` rebuilt. Version **1.2.79 → 1.2.80**.


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new WebSocket-output redactor and agent guidance are a good start, but PR #460 does not yet satisfy #454's old-transcript acceptance: restored IndexedDB scrollback is still written to xterm without redaction, so existing fiction/cartoon terminal transcripts can continue to display previously persisted auth material.

### Findings
- [high] Restored scrollback bypasses the redaction path.
  - File: `app/web/components/TerminalPanel.tsx:331`
  - Suggestion: Apply `redactTerminalSecrets` to `saved` before `term.write`, and if the redacted text differs, persist the redacted scrollback back to IndexedDB so the raw value is not re-saved later. Add a regression test where `loadScrollback` returns a placeholder Authorization/Bearer/passphrase fragment and assert the raw placeholder is never written.

### Decision
Requesting changes because #454 specifically requires opening old terminal transcripts with sensitive values masked; currently only new WebSocket frames are redacted. Live CI `lint-and-typecheck` was still pending when reviewed.

realproject7 (commented):
RE2 APPROVE (pending green CI) — recorded as a comment per the shared-account policy.

No blocking findings. Redaction patterns keep the key/prefix and mask only the value, with value classes excluding whitespace/quotes/&/ESC (so tokens are fully masked but trailing ANSI is never swallowed); covers exactly the ticketed shapes (Authorization: Bearer, standalone Bearer with 12+ min-length so prose 'Bearer' is safe, token=, OWS_PASSPHRASE, --passphrase, quoted passphrase:); idempotent; no-match input unchanged. Display/log-only: wired on the output path only (after the #453 signal check), input/typed commands unredacted so auth/publish/login still work. Persisted scrollback is redacted too — TerminalPanel writes redactTerminalSecrets(e.data) to the term and SerializeAddon persists that buffer to IndexedDB, no raw-secret persist path. No server-side leak vector (terminal.ts never logs raw PTY output — onData only pipes to the WS). Agent guidance added to fiction + cartoon story instructions AND central CLAUDE.md. Tests use placeholders only (#437-aware \x1b escapes) and cover per-pattern + Bearer-prose-negative + ANSI-preserved + idempotent + masks-before-write + guidance-present. Version 1.2.80; dist rebuilt; fiction restore + #453 behavior preserved.

Minor non-blocking (correctly handled): a token split across two streamed frames can slip through per-frame redaction — documented in the lib and mitigated by the agent guidance (defense-in-depth). Clear to merge once CI green + @re1 signs off — completes Batch 40.

realproject7 (commented):
RE2 re-review on fe219b7 — APPROVE (pending green CI). Recorded as a comment per the shared-account policy.

Concurring with @re1's CHANGES_REQUESTED on the prior commit: redaction covered the live WS path + forward-persist but NOT the restore path — createSession wrote loadScrollback's result to xterm unredacted, so transcripts persisted before this feature leaked secrets on open (the ticket's core 'open old transcripts masked' acceptance). My earlier approval missed that.

The fix is correct: the restore path now does redactTerminalSecrets(saved) before term.write, and re-saves the masked copy when it changed (raw value purged from IndexedDB, never re-persisted). Both live (onmessage) and restore are now covered, and the SerializeAddon forward-persist serializes the redacted buffer. New regression test seeds IndexedDB with OWS_PASSPHRASE + Authorization: Bearer placeholders, opens the story, and asserts the raw secrets are never written AND the re-persisted scrollback is [REDACTED] only. Version 1.2.81; dist rebuilt; placeholders only. Clear to merge once CI green + @re1 re-approves.

project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #460 now satisfies #454. The redactor covers new terminal WebSocket output and restored IndexedDB scrollback, and restored raw placeholder auth material is re-saved in redacted form so old transcripts no longer keep exposing it.

### Findings
- No blocking findings.

### Decision
Approved at `fe219b7` (`fe219b76733c1b34ec5dede0165458b27b1989b6`). The previous restored-scrollback finding is fixed: `createSession` redacts `saved` before `term.write` and re-persists the masked copy when changed, with a regression test covering old scrollback. Live CI `lint-and-typecheck` passed and the PR is currently `clean`.

PR #477, opened 2026-06-05
[#470] Stabilize the CLI runtime vs build-time boundary (EPIC #465)

realproject7:
## #470 — Stabilize the CLI runtime vs build-time boundary (EPIC #465)

#469 split the *dependency lists*; **#470 makes the CLI `start` path honour that
split** so an installed package never reaches for build-time tooling it doesn't have.

### Problem
After #469 moved `vite`/`react`/etc. to `devDependencies`, `cmdStart` still ran
`npx vite build` (and `npm install`) whenever `app/web/dist` or `node_modules`
looked absent. On a user's machine that build tooling is no longer installed, so
the fallback would **fetch the entire toolchain from the network — an unexpected
rebuild** on the user's machine, exactly what #470 must prevent.

### Changes
- **`bin/startup-plan.cjs`** — a pure, unit-tested boundary planner. A
  (re)build/install is only allowed in a **source checkout**, detected by `src/`
  (the Next.js app, never in the published `files` allowlist). In an installed
  package, missing `node_modules`/`dist` is a **corrupted install** → clear error
  + reinstall hint, *not* a silent network rebuild.
- **`bin/plotlink-ows.js`** — `cmdStart` uses the planner. Runtime-deps presence
  is probed with `require.resolve` (not a `node_modules/` directory check) so
  hoisted global (`-g`) installs aren't misread as broken. Boundary documented inline.
- **`DEPENDENCIES.md`** — new "Runtime vs build-time boundary" section: prebuilt
  `dist` is shipped; the start path never builds/installs on user machines;
  Prisma generation is deterministic and install-time (`postinstall`).
- **Preflight (#466)** — `bin/startup-plan.cjs` added to `REQUIRED_PACK_FILES`
  (the bin requires it at runtime). Applies the #466/#469 required-set-completeness lesson.

### Safety
- No change to wallet/publish/agent behaviour or the server. The bin's Ctrl+C
  handler is untouched. Source-checkout dev workflow preserved (still auto-builds).
- No generated files shipped to make startup work — relies on the already-committed
  prebuilt `dist`.

### Verification
- `npm run typecheck` — clean
- `npm test` — **1245 passed** (7 new boundary-planner tests)
- `npm run app:build` — clean
- `npm run preflight` (#466) — passes: 0 prod vulns, pack 127 entries / 576KB,
  required runtime contents present (incl. `bin/startup-plan.cjs`), tarball install smoke green
- **Packed-tarball prod-only boot smoke** (acceptance proof): `npm pack` →
  `npm install --omit=dev` (112 pkgs, **vite/react absent**) → booted the server
  from the prebuilt `dist`: `GET /api/auth/status` → `{"configured":false}`,
  `GET /` served the dist `index.html`, **no build/install step**; Prisma
  postinstall generated the client and SQLite worked.
- Lock synced to **1.2.90** (package.json + both lock root fields).

Version 1.2.89 → 1.2.90.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -75,10 +75,38 @@ deps into a separate workspace, e.g. `packages/web`) removes the warning from th
 root install entirely. That's a larger structural change (out of #469's "smallest
 safe change" scope) — tracked here rather than rushed.
 
+## Runtime vs build-time boundary (#470)
+
+#469 split the *dependency lists*; #470 makes the **start path** honour that split
+so an installed CLI never reaches for build-time tooling it doesn't have.
+
+The contract:
+
+- **Runtime deps install with the package; build tooling does not.** Build tooling
+  (`vite`, `tailwindcss`, `react`, …) is `devDependencies`, so it is **absent**
+  from `npm install -g plotlink-ows` / `npx plotlink-ows`.
+- **The web UI is prebuilt and shipped.** `prepublishOnly` runs `app:build`, and
+  `app/web/dist` is committed + packed. The runtime serves the static `dist`; it
+  never builds it.
+- **The start path must not build or `npm install` on a user's machine.**
+  `bin/plotlink-ows.js` start logic (`bin/startup-plan.cjs`, unit-tested) only
+  runs a build/install in a **source checkout** — detected by `src/`, which is
+  never in the published `files` allowlist. In an installed package, missing
+  `node_modules` or `app/web/dist` is treated as a **corrupted install** (clear
+  error + reinstall hint), *not* a trigger to fetch the toolchain from the
+  network. The deps probe uses `require.resolve` (not a `node_modules/` dir
+  check) so hoisted global (`-g`) installs aren't misread as broken.
+- **Prisma generation is deterministic and install-time.** The `postinstall`
+  runs `prisma generate --schema app/prisma/schema.prisma` once at install; the
+  schema is shipped (and preflight-required). The runtime imports the generated
+  `@prisma/client` and never regenerates at start.
+
 ## Guards
 
-`npm run preflight` (#466) verifies `lib/genres.ts` and the other required
-runtime files are in the packed tarball, and a packed-tarball install smoke test
-confirms `npx plotlink-ows` installs and the bin/runtime entrypoints are present.
-A manual prod-only boot test (`npm pack` → install with prod deps only → run the
-server) confirms the CLI starts with `dependencies` alone.
+`npm run preflight` (#466) verifies `lib/genres.ts`, `bin/startup-plan.cjs`, and
+the other required runtime files are in the packed tarball, and a packed-tarball
+install smoke test confirms `npx plotlink-ows` installs and the bin/runtime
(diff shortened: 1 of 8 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #477 satisfies #470. The CLI start path now enforces the runtime/build-time boundary: installed packages use shipped runtime deps and prebuilt app/web/dist, while missing installed-package assets fail as corrupted installs instead of triggering npm install or a Vite rebuild.

### Findings
- No blocking findings.

### Decision
Approved at 7307d1e4320fd19a26a41b838e3cd5d63f5b597d, pending green CI. I reviewed the live PR diff, issue #470, and live checks. The new startup planner is pure and unit-tested for source-checkout versus installed-package cases; cmdStart uses require.resolve for runtime dependency detection to avoid false failures in hoisted/global installs; bin/startup-plan.cjs is added to REQUIRED_PACK_FILES so preflight catches a missing runtime helper; DEPENDENCIES.md documents the boundary; and package.json plus both package-lock root version fields are synced to 1.2.90. Live lint-and-typecheck was still pending when reviewed.

realproject7:
**@re2 APPROVE** — #470 correctly makes the start path honour the #469 dependency split.

What I verified (not just trusting the summary):

**The boundary planner is sound.** `bin/startup-plan.cjs:planStartup` is pure and exhaustively unit-tested (7 cases): installed-package missing-deps and missing-dist both yield a fatal `error` with `install:false/build:false` (NEVER a rebuild), deps-error takes priority when both are missing, and the source-checkout path keeps the dev auto-install/build convenience. Test file is excluded from the pack; `bin/startup-plan.cjs` IS packed.

**The consumer feeds correct facts.**
- `isSourceCheckout` = `src/` exists at `PROJECT_DIR` — confirmed `src/` is NOT in the published `files` allowlist, so an installed package can never false-positive into a rebuild. `PROJECT_DIR = path.dirname(__dirname)` (package-root-relative), so detection is robust to the user's cwd.
- `depsInstalled` = `require.resolve("tsx", { paths: [PROJECT_DIR] })` — resolves up the tree, correctly handling `-g` hoisted installs (the old `node_modules/` dir check would have misread those). `tsx` is the right canary since the server launches via `npx tsx app/server.ts`.
- `distBuilt` = `app/web/dist/index.html` (tightened from the old dir-exists check; matches `REQUIRED_PACK_FILES`).

**Audited the whole start path for other rebuild triggers** — the only `npm install` / `vite build` `execSync` calls (bin lines 252/256) are now both gated behind `plan.install`/`plan.build`. The OWS SDK load is `require`-only (no auto-install). Error path exits 1 with reinstall hints — fails loudly, no silent network fetch.

**#466 required-set lesson applied** — `bin/startup-plan.cjs` added to `REQUIRED_PACK_FILES` + asserted in the hygiene test, since `bin/plotlink-ows.js` requires it at runtime.

Lock synced to 1.2.90; version bump (3rd digit) appropriate. No blocking findings.

Note for @head: CI `lint-and-typecheck` was still pending at review time — please confirm green before merge.

Issue #484, opened 2026-06-05
Fix packed prod-only Prisma schema engine startup failure


realproject7:
Parent: #465
Blocks: #472

## Problem
Final local publish-readiness verification on `main` `1.2.93` still fails in the real packed prod-only startup smoke.

`npm install`, `npm run typecheck`, `npm test`, and `npm run app:build` pass under Node 20.20.2 / npm 10.8.2. `npm audit --omit=dev` reports 0 vulnerabilities. However `npm run preflight` fails at the `Prod-only start smoke (packed tarball install + boot)` stage.

The packed tarball installs with `--omit=dev`, removed deps are absent, and #481's browser auto-open guard is working, but the real bin exits during server startup before `/api/auth/status` is served.

## Reproduction
On local `main` after PR #483:

```bash
source ~/.nvm/nvm.sh && nvm use 20.20.2
npm install
npm run typecheck
npm test
npm run app:build
npm run preflight
```

Observed preflight failure:

```text
=== Prod-only start smoke (packed tarball install + boot) ===
added 69 packages in 7s
  ✓ tarball installed prod-only (--omit=dev, scripts on)
  ✓ removed deps absent from prod install (@aws-sdk/client-s3, react, vite)
  ✗ server did not serve /api/auth/status within 60s (bin exitCode=1)
  ✗ server did not serve the prebuilt web UI at /

--- captured server output ---
Prisma schema loaded from app/prisma/schema.prisma
Datasource "db": SQLite database "local.db" at "file:/.../home/.plotlink-ows/data/local.db"

Error: Schema engine error:

  ✗ Database setup failed (prisma db push).
    schema:   /private/var/.../node_modules/plotlink-ows/app/prisma/schema.prisma
    database: file:~/.plotlink-ows/data/local.db
    reason:   Command failed: /Users/cho/.nvm/versions/node/v20.20.2/bin/node /private/var/.../node_modules/prisma/build/index.js db push --schema /private/var/.../node_modules/plotlink-ows/app/prisma/schema.prisma --skip-generate
```

## Important context
- #479 fixed the earlier `npx prisma` path by resolving the local Prisma CLI and invoking it with `node`.
- #481 fixed the smoke/preflight browser auto-open side effect via `PLOTLINK_OWS_NO_OPEN=1`.
- This is now a deeper Prisma schema-engine/runtime packaging startup failure. Do not revert #479/#481.
[...]

PR #483, opened 2026-06-05
[#481] Skip CLI browser auto-open during release start smoke

realproject7:
## #481 — Disable browser auto-open during packed start smoke

Follow-up for EPIC #465; release-preflight polish after #479 / PR #480. Operator gate #472 stays open.

### Problem
The #479 packed start smoke (`scripts/start-smoke.mjs`, run by `npm run preflight`) boots the real CLI bin, which always scheduled a browser auto-open ~2s after startup. So a non-interactive release check could pop a browser tab/window on the operator machine.

### Fix (narrow, env-guarded)
- **`bin/startup-plan.cjs`** — new pure helper `shouldAutoOpen(env)`: returns `true` by default, `false` only when `PLOTLINK_OWS_NO_OPEN` is exactly `"1"`. (Lives next to `planStartup`, the existing startup-decision module the bin already requires.)
- **`bin/plotlink-ows.js`** — gate the auto-open `setTimeout` on `shouldAutoOpen(process.env)`.
- **`scripts/start-smoke.mjs`** — pass `PLOTLINK_OWS_NO_OPEN=1` into the spawned bin.
- **`bin/startup-plan.test.ts`** — cover default-open, disabled (`"1"`), and other-values (`"0"`/`""`/`"true"` still open).

### Behavior preserved
Normal `npx plotlink-ows` startup auto-opens exactly as before (env var unset). Only the smoke/preflight path is suppressed. No wallet / publish / dashboard / fiction / cartoon / agent-session / web-UI behavior touched.

### Verification (Node 20.20.2 / npm 10.8.2)
- `npm run typecheck` — pass
- `npm test` — pass (1254 tests; +3 for the new guard)
- `npm run app:build` — pass (committed dist unchanged)
- `npm run preflight` — pass, 0 warnings / 0 failures; the start smoke now boots the bin with `PLOTLINK_OWS_NO_OPEN=1` and still serves `/api/auth/status` + `/`.

Version 1.2.92 → 1.2.93.

### Acceptance mapping (#481)
- [x] typecheck / test / app:build / preflight pass
- [x] Normal user startup behavior preserved (default still auto-opens)
- [x] Preflight / start-smoke no longer opens the browser (env guard wired + unit-tested)
- [ ] #472 stays open for the final operator publish-readiness gate; no npm publish

Note: #482 is a duplicate of this ticket (same title) — implementing #481 only, per @head.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #480, opened 2026-06-05
[#479] Fix packed prod-only startup: resolve local Prisma CLI for db push

realproject7:
## #479 — Fix packed prod-only startup failing at Prisma `db push`

Follow-up for EPIC #465; unblocks operator gate #472.

### Root cause
A packed tarball installed with `npm install --omit=dev ./plotlink-ows-*.tgz` exited during startup `db push` before serving `/api/auth/status`. Startup invoked `npx prisma db push`, which resolves against the process **cwd** and can try to **network-download** Prisma when the bin is started from an unexpected cwd (or offline/sealed). That is fragile in a prod-only install.

### Fix
- **`app/lib/prisma-cli.ts`** (new): resolves the locally-installed Prisma CLI (`prisma/build/index.js`) via Node module resolution from the app dir — works whether deps are nested (source checkout) or hoisted (packed/global install). Throws a clear error if `prisma` is missing.
- **`app/server.ts`**: invoke that CLI with `node` instead of `npx prisma` (no cwd ambiguity, no network). `mkdirSync(DATA_DIR)` before `db push` so a fresh install's SQLite parent dir exists. On failure, print a useful, HOME-redacted diagnostic instead of a raw `execFileSync` stack, then exit 1 (no silent masking).
- **`scripts/start-smoke.mjs`** (new): a real packed-tarball **START** smoke — `npm pack`, `npm install --omit=dev` (scripts on, like a real user install), assert `@aws-sdk/client-s3`/`react`/`vite` are absent, boot the bin with a fresh `HOME` + minimal config, and assert the server serves `/api/auth/status` **and** `/`. Captures and prints server output on failure.
- **`scripts/preflight.mjs`**: run the start smoke as a publish gate (step 5). Skippable via `PREFLIGHT_SKIP_START_SMOKE=1`, which warns loudly that a skipped run is not publish-safe.
- **`scripts/package-hygiene.mjs`** + test: require `app/lib/prisma-cli.ts` in the packed contents (server imports it at boot).

The existing pack smoke only checked file presence with `--ignore-scripts`; it could not catch a startup regression. The new start smoke does, and reproduces/fixes the exact #479 failure.

### Boundary preserved
No runtime `@aws-sdk/client-s3`, React, Vite, or web-build deps reintroduced; no user-managed IPFS/S3 credentials. `prisma` was already a runtime dependency (allowlist still 11 pkgs). Source-checkout dev flow (`npm run app:dev`) and the PlotLink API upload/publish flow are unchanged.

[...]

Issue #472, opened 2026-06-04
Operator Gate: final npm publish readiness check, no publish


realproject7:
Parent: #465
Type: Operator Gate

## Goal
Final human/operator readiness check before publishing a cleaned `plotlink-ows` npm package. This ticket does not publish; the operator will handle npm login and publish manually after sign-off.

## Required Inputs
- All code sub-tickets under #465 are merged.
- Latest `main` is pulled locally.
- Node 20.x and npm 10.x are active.

## Checklist
- Run `npm install` and confirm warnings are acceptable/documented.
- Run `npm audit --omit=dev` and confirm remaining advisories are acceptable/documented.
- Run `npm run typecheck`.
- Run `npm test`.
- Run `npm run app:build`.
- Run the package hygiene/preflight command from #466.
- Confirm `npm pack --dry-run --json` package contents are clean.
- Install the packed tarball in a temp directory and smoke-test `plotlink-ows` startup.
- Confirm no secrets/passphrases/wallet data are included in package contents, logs, fixtures, or docs.
- Confirm npm `latest`/dist-tag plan with the operator before publish.

## Acceptance
- Operator posts the final prepublish result and either approves manual npm publish or opens follow-up tickets.
- No `npm publish` is run by QuadWork agents.

realproject7:
#472 final operator gate is currently blocked by #479.

Local final verification after Batch 42 found that the package hygiene work is mostly successful (`npm audit --omit=dev`, typecheck, tests, build, preflight all pass; packed runtime deps exclude `@aws-sdk/client-s3`, React, and Vite), but the real packed-tarball install/start smoke still fails before serving `/api/auth/status`:

```text
Error: Schema engine error:
Error: Command failed: npx prisma db push --schema .../node_modules/plotlink-ows/app/prisma/schema.prisma --skip-generate
```

Keep this operator gate open until #479 is fixed and the packed tarball startup smoke passes under Node 20/npm 10.

realproject7:
Final local publish-readiness verification after Batch 44 / PR #483 is still blocked. #481 is fixed, but 
> plotlink-ows@1.2.93 preflight
> node scripts/preflight.mjs


=== Expected toolchain ===
  engines.node:    20.x
  engines.npm:     10.x
  packageManager:  npm@10.9.8
  running node v24.14.1 / npm 11.11.0
  ! Node v24.14.1 is not 20.x — publish should run on Node 20.x (engines.node).
[...]

PR #485, opened 2026-06-05
[#484] Apply SQLite schema without the native Prisma schema-engine

realproject7:
## #484 — Fix packed prod-only Prisma schema-engine startup failure

Follow-up for EPIC #465; unblocks operator gate #472. Builds on #479 (PR #480) and #481 (PR #483) — does not revert either.

### Root cause
Even after #479/#481, the operator's final verification on `main` 1.2.93 (Node 20.20.2 / npm 10.8.2, macOS arm64) failed in the packed start smoke: startup `prisma db push` exited with an **empty** `Schema engine error:` before `/api/auth/status` served.

`prisma db push` spawns the native **Prisma schema-engine** — a separate, platform-specific binary (`@prisma/engines/schema-engine-*`). That binary fails to start in some packed prod-only installs (empty error on macOS arm64). #479 made the CLI *resolution* robust, but the schema-engine itself is the fragile piece, and it only runs `db push`. (This is why the Linux start smoke went green for #479/#481 — Linux's schema-engine works.)

### Fix — stop using the schema-engine at runtime
The app's normal operation uses the Prisma client's **library query engine** (a different, reliably-present engine — if it failed, the app couldn't serve anything). So we let *it* set up the schema instead of the schema-engine:

- **`app/prisma/schema.sql`** (new, committed): the canonical DDL generated from `schema.prisma`. Regenerate with **`npm run prisma:sql`** (`scripts/gen-schema-sql.mjs`) — that uses the schema-engine at **dev/build time only**, never at user runtime.
- **`app/lib/apply-schema.ts`** (new): `parseSqlStatements` / `makeIdempotent` / `loadSchemaStatements` — reads the DDL and rewrites each `CREATE TABLE/INDEX` to `IF NOT EXISTS` (idempotent on every boot).
- **`app/server.ts`**: connect the client, then apply the DDL via `db.$executeRawUnsafe(...)`. No `prisma db push`, no child-process shell-out for DB setup. Keeps the HOME-redacted diagnostic + exit 1.
- **Removed** now-unused `app/lib/prisma-cli.ts` (+test): the runtime no longer invokes *any* Prisma binary, which strictly supersedes #479's CLI-resolution (it does **not** reintroduce the `npx` network/cwd bug #479 fixed). Pack-hygiene allowlist swaps that entry for `app/lib/apply-schema.ts` + `app/prisma/schema.sql`.

### Proof it removes the schema-engine dependency
Beyond preflight passing, I ran a decisive check: in a packed `--omit=dev` install I **deleted the schema-engine binary** (`@prisma/engines/schema-engine-debian-openssl-3.0.x`, the only one present) and started the bin with a fresh HOME — **the app still booted and served `/api/auth/status`**. The old `prisma db push` path would fail there (that's the operator's exact symptom). Platform-independent fix confirmed.

### Regression coverage
- Idempotency + parse unit tests for `apply-schema`.
- Model→table **sync** test: every `model` in `schema.prisma` has a `CREATE TABLE` in `schema.sql` (catches a schema change that forgot `npm run prisma:sql`).
- **Source contract** test: `app/server.ts` applies `schema.sql` and never runs `db push`/`execSync` — because the Linux start smoke passes either way, this is what locks the fix against a schema-engine regression.

### Boundary preserved
Runtime `dependencies` allowlist unchanged (still 11; `prisma` stays for the postinstall `prisma generate`, which uses the query engine and already works). No React/Vite/AWS/S3/build deps added. No wallet/publish/dashboard/fiction/cartoon/agent-session/web-UI behavior changed.

### Verification (Node 20.20.2 / npm 10.8.2)
- `npm run typecheck` — pass
- `npm test` — pass (1261; +7 net)
- `npm run app:build` — pass (committed dist unchanged)
- `npm run preflight` — pass, 0 warnings / 0 failures (packed start smoke boots + serves `/api/auth/status` + `/`)
- `npm audit --omit=dev` — 0 vulnerabilities
- schema-engine-deleted boot proof — pass

Version 1.2.93 → 1.2.94.

### Acceptance mapping (#484)
- [x] Diagnosed exact schema-engine failure (root cause above)
- [x] Fixed packed install/start path without weakening dependency hygiene
- [x] Focused regression coverage added (sync + idempotency + source contract)
- [x] typecheck / test / app:build / preflight pass; `npm audit --omit=dev` clean
- [ ] #472 stays open until the operator reruns final publish-readiness verification


project7-interns (approved):
## Verdict: APPROVE

### Summary
PR #485 satisfies #484. It removes the runtime dependency on Prisma schema-engine startup by applying committed SQLite DDL through Prisma Client, while preserving the production dependency boundary and packed start-smoke coverage.

### Findings
- No blocking findings.

### Decision
Approved at `b565d52f33ccdc1196d2a7789f5056b571c6cb01`, pending green CI. I reviewed the live PR diff, issue #484, and live checks. The startup path no longer invokes `prisma db push` or shells out for database setup; `app/server.ts` connects Prisma Client, loads `app/prisma/schema.sql`, and applies idempotent `CREATE TABLE/INDEX IF NOT EXISTS` statements through `$executeRawUnsafe`. The packed runtime now explicitly requires `app/lib/apply-schema.ts` and `app/prisma/schema.sql`, while React, Vite, AWS/S3, and other build/web deps remain out of runtime `dependencies`. Regression coverage locks the parser/idempotency behavior, committed DDL model coverage, package contents, and the no-`db push` startup contract. Live `lint-and-typecheck` was still pending when reviewed. Local tests were not run in this checkout.

realproject7:
**@re2 APPROVE** — #484 correctly fixes the schema-engine startup failure. Well-engineered with decisive proof.

Reviewed the PR commit (`b565d52`):
- **Approach is sound.** Replacing `prisma db push` (fragile native schema-engine) with the committed DDL applied via the client's library query engine (`db.$executeRawUnsafe`, per-statement loop, `CREATE … IF NOT EXISTS`) is the right call — the query engine is already required for every app query, so "if the app runs, schema setup runs." The proof (packed `--omit=dev` install with the schema-engine binary *deleted*, app still boots + serves) directly reproduces and clears the operator's macOS arm64 symptom.
- **Ordering is safe.** `initDb()` (just `db.$connect()`, no table reads) is moved before the DDL loop — connecting doesn't require tables, so a fresh DB is fine.
- **schema.sql matches schema.prisma.** Both models present — Session (id/token/createdAt/expiresAt) + Setting (key/value) + the `Session_token_key` UNIQUE index for `token @unique`. Idempotency helpers + the model→table sync test guard against a forgotten `npm run prisma:sql`.
- **Required-set consistency is correct** (the #466/#470 lesson): the now-deleted `app/lib/prisma-cli.ts` is removed from `REQUIRED_PACK_FILES`, and `apply-schema.ts` + `schema.sql` are added — confirmed both pack (857B/2.2kB), tests excluded, prisma-cli gone. The source-contract test (server.ts has `loadSchemaStatements`, no `db push`/`execSync`) locks the fix; confirmed no other `execSync` in server.ts so it won't false-fail.
- **Boundary preserved**: deps still 11 (`prisma` stays for postinstall `generate`); no AWS/React/Vite. Lock + version synced 1.2.94.

Two NON-BLOCKING, forward-looking notes (schema-evolution, not this PR's scope — current Session/Setting schema is unaffected):
1. `CREATE … IF NOT EXISTS` won't ALTER an existing table, so a *future* column addition won't migrate existing users' DBs the way `db push` would. Worth tracking an explicit migration path before any non-additive/seeded schema change.
2. The model→table sync test asserts table-NAME presence, not columns — a column drift (new field in schema.prisma without `prisma:sql`) would slip through. A regen-and-diff or column-level assertion would close that gap later.

No blocking findings.

Note for @head: CI `lint-and-typecheck` was still pending at review time — confirm green before merge.

PR #489, opened 2026-06-07
[#196] Add active OWS wallet selection

realproject7:
## Summary
- Add a centralized active OWS wallet resolver backed by the local `Setting` table, with one-wallet auto-selection and explicit selection required for ambiguous multi-wallet installs.
- Add wallet discovery/switching APIs and UI, plus unique `plotlink-writer-N` wallet creation that makes the new wallet active.
- Route publishing, upload signatures, storyline updates, dashboard balances/royalties, and AI writer registration/binding through the selected active wallet instead of the first `plotlink-writer*` wallet.
- Scope cached AI writer/agent metadata to the active wallet, so switching wallets cannot reuse another wallet's cached agent context.
- Return only non-secret wallet metadata to the UI and tests cover fake multi-wallet switching.

## Verification
- `npm run typecheck`
- `TMPDIR=/home/quadwork/plotlink-ows-dev/.tmp/vitest npm test -- app/lib/active-wallet.test.ts app/routes/settings.test.ts app/routes/publish.test.ts app/routes/upload-cover.test.ts`
- `npm run lint` (passes; existing unrelated warnings remain)
- `npm run app:build`

Fixes #196


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
PR #489 covers most active-wallet routing, but the AI writer/link-status cache still mixes global agent metadata with the currently selected wallet. That breaks issue #196's requirement that dashboard/account/AI writer flows reflect the selected wallet and that switching wallets does not mix context.

### Findings
- [high] `/api/settings/link-status` returns the cached `config.agentId` for whichever wallet is currently active, without proving that cached agent belongs to that wallet. After registering or caching Agent A for Wallet A, switching to Wallet B makes this branch return `linked: true`, Agent A's id, and Wallet B's address before any wallet-specific RPC check runs. The same global config metadata is also included in `/generate-binding`, so binding UI can carry stale agent data across wallet switches.
  - File: `app/routes/settings.ts:207`
  - Suggestion: key cached agent/link metadata by active wallet address/name, or only use the cache when it is explicitly associated with the active wallet. Otherwise fall through to `agentIdByWallet(address)` / ownership checks and update wallet-scoped cache entries. Add a regression test for two fake active wallets where switching from a registered/cached wallet to an unregistered wallet reports unlinked and does not return the first wallet's agent metadata.

### Decision
Request changes at `1fe9d32fec6da6b4bff5093efdfe47abb37d7027`. Live CI `lint-and-typecheck` is green, and the rest of the resolver/API/UI direction looks aligned, but this cache bug is a blocking acceptance-criteria miss for active-wallet switching.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
PR #489 moves the main wallet discovery/signing paths toward an explicit active wallet, but AI writer link status still uses a global agent cache. After switching wallets, the UI/API can report the newly selected wallet as linked to the previously selected wallet's agent, which violates #196's requirement that account/AI writer views reflect the selected wallet and not mix context across switches.

### Findings
- [high] `link-status` trusts the global cached `config.agentId` before checking whether it belongs to the active wallet.
  - File: `app/routes/settings.ts:207`
  - Details: `register-agent` writes `agentId` into a shared `config.json` cache, and `link-status` immediately returns `linked: true` whenever that cache exists, using the current active wallet address in the response. With wallet A registered/cached, switching to wallet B will still return wallet A's `agentId` as linked for wallet B without calling `agentIdByWallet(B)`. This directly mixes AI writer/account state across active-wallet switches.
  - Suggestion: Scope cached agent metadata by active wallet identity/address, or only use the cache when it records the same active wallet address. Otherwise fall through to the on-chain `agentIdByWallet`/ownership checks and update the cache under that wallet-specific key.

### Decision
Requesting changes. I reviewed live PR #489 at `1fe9d32fec6da6b4bff5093efdfe47abb37d7027`, issue #196, and live checks. CI was still pending (`lint-and-typecheck`) at review time.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested active-wallet fix is in place at `d9ca926cee7a4b67a56b3f25782080b13b86a982`. Cached AI writer/agent metadata is now scoped to the active wallet, and stale Wallet A metadata is not reused after switching to Wallet B.

### Findings
- No blocking findings.

### Decision
Approved. I re-reviewed the live PR diff, issue #196, and the current settings/cache changes. `app/routes/settings.ts` now validates cached agent metadata against active wallet address/id/name before using it, writes wallet-scoped metadata after registration/RPC discovery, and `app/routes/settings.test.ts` covers the two-fake-wallet stale-cache regression for both `link-status` and `generate-binding`. CI was still pending when I checked the current commit.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested active-wallet cache fix is in place. AI writer agent metadata is now scoped to the active wallet identity, and stale Wallet A agent metadata is no longer reused after switching to Wallet B.

### Findings
- No blocking findings.

### Decision
Approved at `d9ca926cee7a4b67a56b3f25782080b13b86a982`, pending the new live CI run. I reviewed the updated `settings.ts` logic and the added `settings.test.ts` regression: cached agent metadata is trusted only when it matches the active wallet address/id/name, otherwise `link-status` falls through to wallet-specific on-chain checks, and `generate-binding` omits stale cached agent fields for the newly active wallet. This resolves my prior request-change finding for #196.

